PR #9, opened 2014-12-25
Switch off of Elasticsearch as main storage.

polesye:
It was decided that ElasticSearch as a storage is not reliable and tried enough, so we should recreate the app with MySQL as main data storage.

Should support ES 0.90

### DIFF
--- a/notesapi/management/commands/create_index.py
+++ b/notesapi/management/commands/create_index.py
@@ -1,9 +1,0 @@
-from django.core.management.base import BaseCommand
-from annotator.annotation import Annotation
-
-
-class Command(BaseCommand):
-    help = 'Creates the mapping in the index.'
-
-    def handle(self, *args, **options):
-        Annotation.create_all()
--- a/notesapi/v1/models.py
+++ b/notesapi/v1/models.py
@@ -1,0 +1,57 @@
+import json
+from django.db import models
+from django.core.exceptions import ValidationError
+
+
+class Note(models.Model):
+    """
+    Annotation model.
+    """
+    user_id = models.CharField(max_length=255, db_index=True, help_text="Anonymized user id, not course specific")
+    course_id = models.CharField(max_length=255, db_index=True)
+    usage_id = models.CharField(max_length=255, help_text="ID of XBlock where the text comes from")
+    quote = models.TextField(default="")
+    text = models.TextField(default="", help_text="Student's thoughts on the quote")
+    ranges = models.TextField(help_text="JSON, describes position of quote in the source text")
+    created = models.DateTimeField(auto_now_add=True)
+    updated = models.DateTimeField(auto_now=True)
+
+    @classmethod
+    def create(cls, note_dict):
+        """
+        Create the note object.
+        """
+        if not isinstance(note_dict, dict):
+            raise ValidationError('Note must be a dictionary.')
+
+        if len(note_dict) == 0:
+            raise ValidationError('Note must have a body.')
+
+        ranges = note_dict.get('ranges', list())
+
+        if len(ranges) < 1:
+            raise ValidationError('Note must contain at least one range.')
+
+        note_dict['ranges'] = json.dumps(ranges)
+        note_dict['user_id'] = note_dict.pop('user', None)
+
+        return cls(**note_dict)
+
+    def as_dict(self):
+        """
+        Returns the note object as a dictionary.
+        """
+        created = self.created.isoformat() if self.created else None
+        updated = self.updated.isoformat() if self.updated else None
+
+        return {
+            'id': str(self.pk),
+            'user': self.user_id,
+            'course_id': self.course_id,
+            'usage_id': self.usage_id,
+            'text': self.text,
+            'quote': self.quote,
+            'ranges': json.loads(self.ranges),
+            'created': created,
+            'updated': updated,
+        }
--- a/notesapi/v1/tests/test_models.py
+++ b/notesapi/v1/tests/test_models.py
@@ -1,0 +1,49 @@
+from unittest import TestCase
+from notesapi.v1.models import Note
+from django.core.exceptions import ValidationError
+
+
+class NoteTest(TestCase):
+    def setUp(self):
+
+        self.note_dict = {
+            "user": u"test_user_id",
+            "usage_id": u"i4x://org/course/html/52aa9816425a4ce98a07625b8cb70811",
+            "course_id": u"org/course/run",
+            "text": u"test note text",
+            "quote": u"test note quote",
+            "ranges": [
+                {
+                    "start": u"/p[1]",
+                    "end": u"/p[1]",
+                    "startOffset": 0,
+                    "endOffset": 10,
+                }
+            ],
+        }
+
+    def test_create_valid_note(self):
+        note = Note.create(self.note_dict.copy())
+        note.save()
+
+        result_note = note.as_dict()
+        del result_note['id']
+        del result_note['created']
+        del result_note['updated']
+
+        self.assertEqual(result_note, self.note_dict)
+
+    def test_create_invalid_note(self):
+        note = Note()
+        for empty_type in (None, '', []):
+            with self.assertRaises(ValidationError):
+                note.create(empty_type)
+
+    def test_must_have_fields_create(self):
+        for field in ['user', 'usage_id', 'course_id', 'ranges']:
+            payload = self.note_dict.copy()
+            payload.pop(field)
+
+            with self.assertRaises(ValidationError):
+                note = Note.create(payload)
+                note.full_clean()
--- a/notesapi/v1/tests/test_views.py
+++ b/notesapi/v1/tests/test_views.py
@@ -1,31 +1,35 @@
+# -*- coding: utf-8 -*-
 import jwt
+import unittest
 from calendar import timegm
 from datetime import datetime, timedelta
 from mock import patch
-import unittest
 
+from django.core.management import call_command
 from django.core.urlresolvers import reverse
 from django.conf import settings
+from django.http import QueryDict
 
 from rest_framework import status
 from rest_framework.test import APITestCase
 
-from annotator import es, auth
-from annotator.annotation import Annotation
 from .helpers import get_id_token
 
-TEST_USER = "test-user-id"
+TEST_USER = "test_user_id"
+
+if not settings.ES_DISABLED:
+    import haystack
+else:
+    def call_command(*args, **kwargs):
+        pass
+
 
 class BaseAnnotationViewTests(APITestCase):
     """
     Abstract class for testing annotation views.
     """
     def setUp(self):
-        assert Annotation.es.host == settings.ELASTICSEARCH_URL
-        assert Annotation.es.index == settings.ELASTICSEARCH_INDEX
-
-        Annotation.create_all()
-        es.conn.cluster.health(wait_for_status='yellow')
+        call_command('clear_index', interactive=False)
 
         token = get_id_token(TEST_USER)
         self.client.credentials(HTTP_X_ANNOTATOR_AUTH_TOKEN=token)
@@ -47,66 +51,36 @@ class BaseAnnotationViewTests(APITestCase):
             ],
         }
 
-        self.expected_note = {
-            "created": "2014-11-26T00:00:00+00:00",
-            "updated": "2014-11-26T00:00:00+00:00",
-            "user": TEST_USER,
-            "usage_id": "test-usage-id",
-            "course_id": "test-course-id",
-            "text": "test note text",
-            "quote": "test note quote",
-            "ranges": [
-                {
-                    "start": "/p[1]",
-                    "end": "/p[1]",
-                    "startOffset": 0,
-                    "endOffset": 10,
-                }
-            ],
-            "permissions": {"read": ["group:__consumer__"]},
-        }
-
-    def tearDown(self):
-        Annotation.drop_all()
-
-    def _create_annotation(self, refresh=True, **kwargs):
+    def _create_annotation(self, **kwargs):
         """
-        Create annotation directly in elasticsearch.
+        Create annotation
         """
-        opts = {
-            'user': TEST_USER,
-        }
+        opts = self.payload.copy()
         opts.update(kwargs)
-        annotation = Annotation(**opts)
-        annotation.save(refresh=refresh)
-        return annotation
+        url = reverse('api:v1:annotations')
+        response = self.client.post(url, opts, format='json')
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        return response.data.copy()
 
     def _get_annotation(self, annotation_id):
         """
-        Fetch annotation directly from elasticsearch.
+        Fetch annotation
         """
-        return Annotation.fetch(annotation_id)
-
-    def _get_search_results(self, qs=''):
-        """
-        Helper for search method.
-        """
-        url = reverse('api:v1:annotations_search') + '?user=' + TEST_USER + '&{}'.format(qs)
-        result = self.client.get(url)
-        return result.data
+        url = reverse('api:v1:annotations_detail', kwargs={'annotation_id': annotation_id})
+        response = self.client.get(url, self.headers)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        return response.data
 
 
-class AnnotationViewTests(BaseAnnotationViewTests):
+class AnnotationListViewTests(BaseAnnotationViewTests):
     """
-    Test annotation views, checking permissions
+    Test annotation creation and listing by user and course
     """
 
-    @patch('annotator.elasticsearch.datetime')
-    def test_create_note(self, mock_datetime):
+    def test_create_note(self):
         """
         Ensure we can create a new note.
         """
-        mock_datetime.datetime.now.return_value.isoformat.return_value = "2014-11-26T00:00:00+00:00"
 
         url = reverse('api:v1:annotations')
         response = self.client.post(url, self.payload, format='json')
@@ -115,8 +89,12 @@ class AnnotationViewTests(BaseAnnotationViewTests):
 
         annotation = response.data.copy()
         self.assertIn('id', annotation)
-        annotation.pop('id')
-        self.assertEqual(annotation, self.expected_note)
+        self.assertEqual(type(annotation['id']), str)
+        del annotation['id']
+        del annotation['updated']
+        del annotation['created']
+
+        self.assertEqual(annotation, self.payload)
 
         expected_location = '/api/v1/annotations/{0}'.format(response.data['id'])
         self.assertTrue(
@@ -130,7 +108,7 @@ class AnnotationViewTests(BaseAnnotationViewTests):
         """
         Test if annotation 'created' field is not used by API.
         """
-        self.payload['created'] = 'abc'
+        self.payload['created'] = '2015-01-05T11:46:58.837059+00:00'
         response = self.client.post(reverse('api:v1:annotations'), self.payload, format='json')
         self.assertEqual(response.status_code, status.HTTP_201_CREATED)
 
@@ -141,9 +119,7 @@ class AnnotationViewTests(BaseAnnotationViewTests):
         """
         Test if annotation 'updated' field is not used by API.
         """
-        self.payload['updated'] = 'abc'
-        payload = self.payload
-        payload.update(self.headers)
+        self.payload['updated'] = '2015-01-05T11:46:58.837059+00:00'
         response = self.client.post(reverse('api:v1:annotations'), self.payload, format='json')
         self.assertEqual(response.status_code, status.HTTP_201_CREATED)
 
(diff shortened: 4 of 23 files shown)


notesapi/management/commands/create_index.py
```diff
@@ -1,9 +1,34 @@
...
+    Indexing and mapping commands.
+    """
+    help = 'Creates index and the mapping.'
+    option_list = BaseCommand.option_list + (
+        make_option(
+            '--drop',
+            action='store_true',
+            dest='drop',
+            default=False,
+            help='Recreate index'),
```
olmar: > value of help seems wrong.

with drop option it will delete existing index and create a new one.

notesapi/management/commands/create_index.py
```diff
@@ -1,9 +1,34 @@
...
+            # drop existing
+            get_es().indices.delete(index=settings.ES_INDEXES['default'], ignore=404)
+        get_es().indices.create(
+            index=settings.ES_INDEXES['default'],
+            body={
+                'mappings': {
+                    NoteMappingType.get_mapping_type_name(): NoteMappingType.get_mapping()
+                }
+            },
+            ignore=400  # ignore when present
```
olmar: > maybe we should not ignore 400 when drop=True?

I think it is better not to have ignores at all.
I removed 400 and 404 code ignores from this command.

notesserver/settings/logger.py
```diff
@@ -0,0 +1,108 @@
...
+                'stream': sys.stdout,
+            },
+        },
+        'loggers': {
+            'django': {
+                'handlers': handlers,
+                'propagate': True,
+                'level': 'INFO'
+            },
+            "elasticsearch.trace": {
```
olmar: @e0d "elasticsearch.trace" is the logger from elasticsearch-py client for Elasticsearch. Not sure if it possible to rename it.

 From [here](http://elasticutils.readthedocs.org/en/latest/debugging.html):
> elasticsearch-py logs to the elasticsearch and elasticsearch.trace loggers using the Python logging module. 
If you configure elasticsearch.trace to show INFO-level messages, then it’ll show the requests in curl form, responses if you enable DEBUG.

notesserver/settings/yaml_config.py
```diff
@@ -1,8 +1,9 @@
 import yaml
-import annotator
-from annotator import es
 
 from .common import *  # pylint: disable=unused-wildcard-import, wildcard-import
+from notesserver.settings.logger import get_logger_config
+
+LOGGING = get_logger_config()
```
olmar: > Should this be in common and overriden in dev?

Fixed.

notesserver/views.py
```diff
@@ -28,7 +27,7 @@ def heartbeat(request):  # pylint: disable=unused-argument
     """
     ElasticSearch is reachable and ready to handle requests.
     """
-    if es.conn.ping():
+    if get_es().ping():
```
olmar: > We should also be checkong MySQL now.

Added.

notesserver/settings/common.py
```diff
@@ -73,17 +74,17 @@
...
             'handlers': ['console'],
             'level': 'DEBUG',
             'propagate': True
         },
-        'elasticsearch': {
+        'elasticsearch.trace': {
             'handlers': ['console'],
-            'level': 'WARNING',
-            'propagate': True
+            'level': 'DEBUG',
```
olmar: > it seems worrisome to me to have loggers default to noisy levels, though I'm not sure if it's really a problem - @e0d ?

Logging is updated.

notesapi/v1/tests/test_models.py
```diff
@@ -0,0 +1,78 @@
...
+    def test_extract_document(self):
+        note = Note()
+        note.clean(self.note)
+        note.save()
+        self.assertEqual(NoteMappingType.extract_document(note.id), note.as_dict())
+
+    def test_get_model(self):
+        self.assertIsInstance(NoteMappingType.get_model()(), Note)
+
+    def test_get_mapping(self):
```
olmar: > it's great how thorough this test case is, but this test in particular seems low in value, while making extra work for the maintainer.

Removed test.

notesapi/v1/models.py
```diff
@@ -0,0 +1,145 @@
...
+    user_id = models.CharField(max_length=255, help_text="Anonymized user id, not course specific")
+    course_id = models.CharField(max_length=255)
+    usage_id = models.CharField(max_length=255, help_text="ID of XBlock where the text comes from")
+    quote = models.TextField(default="")
+    text = models.TextField(default="", help_text="Student's thoughts on the quote")
+    ranges = models.TextField(default="", help_text="JSON, describes position of quote in the source text")
+    created = models.DateTimeField(auto_now_add=True)
+    updated = models.DateTimeField(auto_now=True)
+
+    def clean(self, note):
```
jimabramson: my [earlier comment concerning with this API](https://github.com/edx/edx-notes-api/pull/9#discussion_r22374671) still stands, any plans to address it?

notesapi/v1/tests/test_views.py
```diff
@@ -47,51 +44,66 @@ def setUp(self):
...
+    @classmethod
+    def tearDownClass(cls):
+        """
+        deletes the test index
+        """
+        get_es().indices.delete(index=settings.ES_INDEXES['default'])
+        get_es().indices.refresh()
 
-    def _create_annotation(self, refresh=True, **kwargs):
+    def _create_annotation(self, **kwargs):
```
jimabramson: FYI - [factory boy](http://factoryboy.readthedocs.org/en/latest/) is a great tool for doing what this method and `self.payload` are doing - here's an [example](https://github.com/edx/edx-platform/blob/master/common/test/acceptance/fixtures/discussion.py#L14-L64) of using it to generate dicts from templates in forums bok-choy tests

notesapi/v1/tests/test_views.py
```diff
@@ -126,6 +139,29 @@ def test_create_note(self, mock_datetime):
...
+        """
+        url = reverse('api:v1:annotations')
+        response = self.client.post(url, self.payload, format='json')
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        Note.objects.get(id=response.data['id'])
+        self.assertEqual(note_searcher.filter(id=response.data['id']).count(), 0)
+
+    def test_delete_es_disabled(self):
+        """
+        Ensure we can create note in database when elasticsearch is disabled.
```
jimabramson: s/create/delete/

notesapi/v1/tests/test_views.py
```diff
@@ -126,6 +139,29 @@ def test_create_note(self, mock_datetime):
 
         self.assertEqual(response.data['user'], TEST_USER)
 
+    @patch('django.conf.settings.ES_DISABLED', True)
```
jimabramson: i'm glad this support is being tested...under what scenarios would we expect ES to be disabled by configuration?

In the event of ES service interruptions or cluster health issues in production, should we be able to fall back gracefully to this state, so that end users can still work?

notesapi/v1/views.py
```diff
@@ -70,17 +72,19 @@ def post(self, *args, **kwargs):  # pylint: disable=unused-argument
...
             return Response(status=status.HTTP_400_BAD_REQUEST)
 
-        filtered_payload = _filter_input(self.request.DATA, CREATE_FILTER_FIELDS)
+        note = Note()
 
-        if len(filtered_payload) == 0:
+        try:
+            note.clean(self.request.DATA)
+        except ValidationError as error:
+            log.debug(error)
```
jimabramson: `log.debug(some_message, exc_info=True)` might format better, not sure.

notesapi/v1/tests/test_views.py
```diff
@@ -101,12 +113,10 @@ class AnnotationViewTests(BaseAnnotationViewTests):
     Test annotation views, checking permissions
```
jimabramson: I recognize that it's not new in this PR, and it's not a top priority, but I suggest it would be better to group the tests into three separate test case classes, each corresponding to the view class being tested.

notesapi/v1/tests/test_views.py
```diff
@@ -47,51 +44,66 @@ def setUp(self):
...
-                    "startOffset": 0,
-                    "endOffset": 10,
+    def tearDown(self):
+        for note_id in note_searcher.all().values_list('id'):
+            get_es().delete(
+                index=settings.ES_INDEXES['default'],
+                doc_type=NoteMappingType.get_mapping_type_name(),
+                id=note_id[0][0]
+            )
+        get_es().indices.refresh()
```
jimabramson: would it better to only refresh the specific index being used?  (same question goes for all the occurrences in this PR, and i suspect it would make sense to implement a convenience method to do so)

notesapi/v1/views.py
```diff
@@ -55,11 +56,12 @@ def get(self, *args, **kwargs):  # pylint: disable=unused-argument
         """
         Get a list of all annotations.
         """
-        self.kwargs['query'] = self.request.QUERY_PARAMS.dict()
-
-        annotations = Annotation.search(**kwargs)
+        params = self.request.QUERY_PARAMS.dict()
+        results = NoteMappingType.process_result(
```
jimabramson: sorry i only just noticed this, but it looks like List and Detail `get` methods are fetching notes from ES instead of the database.  I don't think we want this - my assumption would have been that we used the database as a primary source of truth.  Let me know if I'm misunderstanding what's going on here.

notesapi/v1/models.py
```diff
@@ -0,0 +1,145 @@
...
+from django.dispatch import receiver
+from elasticutils.contrib.django import Indexable, MappingType
+
+
+
+class Note(models.Model):
+    """
+    Annotation model.
+    """
+    # no indexes here because retrieval is taken care of by ElasticSearch
```
jimabramson: related to my comment about List and Detail `get` methods, I do think it would make sense to add database indexes for the most commonly anticipated access patterns - `user_id` + `course_id` is mainly what I'm thinking.

notesapi/v1/models.py
```diff
@@ -0,0 +1,141 @@
...
+    user_id = models.CharField(max_length=255)
+    course_id = models.CharField(max_length=255)
+    usage_id = models.CharField(max_length=255)
+    text = models.TextField(default="")
+    quote = models.TextField(default="")
+    ranges = models.TextField(default="")
+    created = models.DateTimeField(auto_now_add=True)
+    updated = models.DateTimeField(auto_now=True)
+
+    def clean(self, note):
```
olmar: > this deviates from the Model.clean definition (which takes no arguments and operates on instance data in-place) - I think it would be best to override the classmethod Model.create to populate the model from dict values, and then use Model.clean only if needed to provide custom validation that the default implementation doesn't provide.

Good idea! Added `Model.create`.

notesapi/v1/models.py
olmar: > my earlier comment concerning with this API still stands, any plans to address it?

Addressed.

notesapi/v1/views.py
olmar: > log.debug(some_message, exc_info=True) might format better, not sure.

Addressed.

notesapi/v1/views.py
olmar: > sorry i only just noticed this, but it looks like List and Detail get methods are fetching notes from ES instead of the database. I don't think we want this - my assumption would have been that we used the database as a primary source of truth. Let me know if I'm misunderstanding what's going on here.

Addressed. Fetching notes from ES was done as it was assumption to use mysql only when storing/udating/deleting notes. Now List and Detail ``get`` methods use database.

notesapi/v1/models.py
olmar: > related to my comment about List and Detail get methods, I do think it would make sense to add database indexes for the most commonly anticipated access patterns - user_id + course_id is mainly what I'm thinking.

Added.

notesapi/v1/tests/test_views.py
olmar: > s/create/delete/

Fixed.

notesapi/v1/models.py
```diff
@@ -0,0 +1,141 @@
...
+                    if k != 'ranges' and isinstance(v, list) and len(v) > 0:
+                        data[i][k] = v[0]
+
+                    # Substitute the value of text field by highlighted result.
+                    if len(item.es_meta.highlight) and k == 'text':
+                        data[i][k] = item.es_meta.highlight['text'][0]
+
+        return data
+
+note_searcher = NoteMappingType.search()
```
tymofij: That is an object which will be searching through NoteMappingType documents, documented at http://elasticutils.readthedocs.org/en/latest/django_api.html?highlight=search#elasticutils.contrib.django.MappingType.search 

similar to django's SomeModel.objects

notesapi/v1/tests/test_views.py
tymofij: yes, and I have added third configuration mode to run our tests with: no ES. In this mode search would work slower (full-scan), less smart about analyzing phrazes and does not do highlighting. All other tasks (creation, modifying, searching by user_id, course_id, usage_id) work.

notesapi/v1/views.py
```diff
@@ -19,31 +21,25 @@ class AnnotationSearchView(APIView):
...
+        for field in ('text', 'quote'):
+            if field in params:
+                params[field + "__match"] = params[field]
+                del params[field]
+        if params.get('highlight'):
+            params.pop('highlight')
+            results = NoteMappingType.process_result(
+                list(
+                    note_searcher.query(**params).order_by("-created").values_dict("_source") \
+                        .highlight("text", pre_tags=['<span>'], post_tags=['</span>'])
```
tymofij: it does not, and with recent switch to `haystack` that is even more complicated. (haystack does not support it)

notesapi/v1/tests/test_views.py
tymofij: nice idea, done

notesapi/v1/tests/test_views.py
tymofij: with recent switch to `haystack` we just use convenient index recreation commands for that.

tymofij:
@jimabramson you comments were addressed, please take a look.

notesapi/v1/models.py
```diff
@@ -0,0 +1,71 @@
...
+        self.range_start = ranges[0]['start']
+        self.range_start_offset = ranges[0]['startOffset']
+        self.range_end = ranges[0]['end']
+        self.range_end_offset = ranges[0]['endOffset']
+
+    def as_dict(self):
+        """
+        Returns the note object as a dictionary.
+        """
+        created = self.created.isoformat() if self.created else None
```
jimabramson: I agree with @polesye's comment here...is it needed for tests or something?

tymofij: just for completeness so this method would return meaningful dict for not-yet-saved None

notesapi/v1/tests/test_views.py
```diff
@@ -1,31 +1,37 @@
...
 
-from annotator import es, auth
-from annotator.annotation import Annotation
 from .helpers import get_id_token
+from notesapi.v1.models import Note
+
+TEST_USER = "test_user_id"
+
+if not settings.ES_DISABLED:
+    import haystack
```
jimabramson: can we add something to the README about when an operator would choose to run with this setting?  again, it's good that this flexibility exists, but I've always assumed it would be needed under exceptional circumstances (ES having problems) vs. as a deliberate configuration.

notesapi/v1/tests/test_views.py
```diff
@@ -47,66 +53,38 @@ def setUp(self):
...
-        }
+        opts = self.payload.copy()
         opts.update(kwargs)
-        annotation = Annotation(**opts)
-        annotation.save(refresh=refresh)
-        return annotation
+        url = reverse('api:v1:annotations')
+        response = self.client.post(url, opts, format='json')
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        call_command('update_index')
```
jimabramson: i'm confused why this is being called every time a note is created in the test - shouldn't the new notes get indexed by the view code automatically?

notesapi/v1/tests/test_views.py
```diff
@@ -47,66 +53,38 @@ def setUp(self):
...
-
-    def _get_search_results(self, qs=''):
+        Fetch annotation
         """
-        Helper for search method.
-        """
-        url = reverse('api:v1:annotations_search') + '?user=' + TEST_USER + '&{}'.format(qs)
-        result = self.client.get(url)
-        return result.data
+        call_command('update_index')
```
jimabramson: same question about the mgmt command here - doesn't seem like a valid test if we're constantly re-indexing everything.

notesapi/v1/tests/test_views.py
```diff
@@ -199,33 +262,48 @@ def test_update(self):
...
-        self._create_annotation(text=u"Foo", id='123', created='2014-10-10')
-        payload = {'id': '123', 'text': 'Bar'}
+        data = self._create_annotation(text=u"Foo")
+        payload = self.payload.copy()
+        payload.update({'id': data['id'], 'text': 'Bar'})
         payload.update(self.headers)
-        url = reverse('api:v1:annotations_detail', kwargs={'annotation_id': 123})
+        url = reverse('api:v1:annotations_detail', kwargs={'annotation_id': data['id']})
         response = self.client.put(url, payload, format='json')
+        call_command('update_index')
```
jimabramson: see above....

notesapi/v1/tests/test_views.py
```diff
@@ -274,121 +355,136 @@ def test_delete_notfound(self):
         response = self.client.delete(url, self.headers)
         self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND, "response should be 404 NOT FOUND")
 
+
+class AnnotationSearchlViewTests(BaseAnnotationViewTests):
```
jimabramson: please fix typo

notesapi/v1/tests/test_views.py
tymofij: removed.

notesapi/v1/tests/test_views.py
tymofij: typo fixed

jimabramson: could we add a test that ensures that recently-deleted notes don't turn up in search results?

notesapi/v1/tests/test_views.py
tymofij: Added the note in README

notesapi/v1/tests/test_views.py
```diff
@@ -274,121 +355,136 @@ def test_delete_notfound(self):
...
-        response = self.client.get(url, self.headers)
-        self.assertEqual(response.status_code, status.HTTP_200_OK)
-        self.assertEqual(len(response.data), 5, "five annotations should be returned in response")
+        results = self._get_search_results(course_id="b")
+        self.assertEqual(results['total'], 1)
+        self.assertEqual(results['rows'][0]['text'], u'Third note')
 
 
 @patch('django.conf.settings.DISABLE_TOKEN_CHECK', True)
 class AllowAllAnnotationViewTests(BaseAnnotationViewTests):
```
jimabramson: are there any tests that, without tokens disabled, missing or incorrect tokens result in a 401?  just took a quick look and couldn't find anything.  also - is it better to have this test in notesserver?

notesapi/v1/tests/test_views.py
tymofij: done, 27dfc20ed06146c416880cd7b8ae1e8edbcdd20a

notesapi/v1/tests/test_views.py
tymofij: no, for all that is wrong with tokens we send 403. As for this one - it is related to a particular note-saving issue, so I think it is more fit in API tests.

notesapi/v1/views.py
```diff
@@ -1,49 +1,86 @@
...
+        return [note.as_dict() for note in query]
 
-        return Response({'total': total, 'rows': results})
+    def get_from_es(self, *args, **kwargs):  # pylint: disable=unused-argument
+        """
+        Search annotations in ElasticSearch
+        """
+        params = self.request.QUERY_PARAMS.dict()
+        query = SearchQuerySet().models(Note).filter(
+            **{f: v for (f, v) in params.items() if f in ('user', 'course_id', 'usage_id', 'text')}
```
jimabramson: a user should only be able to search their own notes, rather than another users' - is that enforced?

tymofij: yes, by checking IDToken's `user`. See https://github.com/edx/edx-notes-api/blob/oleg/models/notesapi/v1/permissions.py#L46

notesserver/views.py
```diff
@@ -1,13 +1,31 @@
...
+    from haystack import connections
+
+    def get_es():
+        return connections['default'].get_backend().conn
+else:
+    from mock import Mock
+
+    def get_es():
+        return Mock(
+            ping=lambda: False,
```
jimabramson: should this be True instead?  if ES_DISABLED is set deliberately, the heartbeat will always return failures which probably isn't desired since it's not actually a problem.  alternatively, maybe it's better if the es check is skipped entirely in the heartbeat, since it's N/A under that configuration.

notesapi/v1/views.py
tymofij: and tested in https://github.com/edx/edx-notes-api/blob/oleg/models/notesapi/v1/tests/test_views.py#L570

notesapi/v1/models.py
jimabramson: ah, got it

notesapi/v1/views.py
jimabramson: ah thanks, now i see.  so every request must have a user id param, and it must always match the token's, correct?  (other case to worry about is a user fetching notes for all users, by omitting the parameter from the request.)

notesapi/v1/tests/test_views.py
jimabramson: cool

jimabramson:
@tymofij @olmar thanks a lot for hard work on this.  done reviewing - looks great, and will thumb it up once remaining comments addressed.

notesapi/v1/tests/test_views.py
tymofij: all call_command('update_index') were removed from tests.

notesapi/v1/tests/test_views.py
tymofij: all call_command('update_index') were removed from tests.

notesserver/views.py
olmar: > should this be True instead? if ES_DISABLED is set deliberately, the heartbeat will always return failures which probably isn't desired since it's not actually a problem. alternatively, maybe it's better if the es check is skipped entirely in the heartbeat, since it's N/A under that configuration.

I like having es check skipped when ES_DISABLED. In such configuration es is not available at all. Nothing to check in test endoints except database.
Addressed here: https://github.com/edx/edx-notes-api/commit/53d568067b0195488e58e0cfbd258c0b9d70b7d7

notesapi/v1/views.py
tymofij: yes. And there is a check for `user` to always be present in the request https://github.com/edx/edx-notes-api/blob/oleg/models/notesapi/v1/permissions.py#L57, and I just added one more test for that 1202e6b85198a6bf7807c8d5bb0256e3185abd18

tymofij:
@jimabramson all comments have been addressed.

jimabramson:
:+1: huzzah!

tymofij:
@olmar please squash and merge.

PR #12, opened 2015-01-05
E0d/additional requirements

e0d:
PR against Oleg's branch since that's where the work seems to be happening.  We need MySQL Python and Gunicorn to deploy.


tymofij:
I guess there is a mistake, as this one is actually against master?

e0d:
Yeah, will fix, expected github to figure it out since the parent was Oleg's branch.

PR #14, opened 2015-01-07
/ operator requires a path, fails on two string

e0d:


tymofij:
hmm, so `/` is used inside yml file?

e0d:
No:

```
In [2]: with open("/home/edward" / ".bashrc"):
    pass
   ...: 
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-2de0426c7ad5> in <module>()
----> 1 with open("/home/edward" / ".bashrc"):
      2     pass

TypeError: unsupported operand type(s) for /: 'str' and 'str'

In [3]: from path import path

In [4]: with open(path("/home/edward") / ".bashrc"):
    pass
   ...: 

```

PR #16, opened 2015-01-12
Missed import

e0d:


tymofij:
I guess we need at least basic test for yaml_config.

PR #17, opened 2015-01-12
Allow reindexing only recent records.

olmar:
Implement ``get_updated_field`` in order to use ``update_index`` command with the start and end dates for indexing within.
Issue: https://openedx.atlassian.net/browse/TNL-1106

More info:
* [SearchIndex.get_updated_field(self)](http://django-haystack.readthedocs.org/en/latest/searchindex_api.html?highlight=get_updated_field#SearchIndex.get_updated_field)
* [haystack ``update_index`` command](http://django-haystack.readthedocs.org/en/latest/management_commands.html#update-index)

Please review: @tymofij 
FYI: @e0d @jimabramson

### DIFF
--- a/notesapi/v1/tests/test_update_index.py
+++ b/notesapi/v1/tests/test_update_index.py
@@ -1,0 +1,79 @@
+import factory
+import datetime
+from django.conf import settings
+from unittest import skipIf
+from django.core.management import call_command
+from django.db.models import signals
+from django.core.urlresolvers import reverse
+from .test_views import BaseAnnotationViewTests
+
+
+@skipIf(settings.ES_DISABLED, "Do not test if Elasticsearch service is disabled.")
+class UpdateIndexTest(BaseAnnotationViewTests):
+    """
+    Tests for update index command.
+    """
+
+    @factory.django.mute_signals(signals.post_save)
+    def test_create(self):
+        """
+        Ensure we can update index with created notes within specific
+        period of time.
+        """
+        start = datetime.datetime.now()
+        self._create_annotation(text=u'First note')
+
+        second_start = datetime.datetime.now()
+        self._create_annotation(text=u'Second note')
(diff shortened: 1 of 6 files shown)


tymofij:
I would like to see test here, something that disabled post_save signals, then adds few notes, makes sure they are not indexed and issues update_index command to index several of them by time range.

jimabramson:
Agree with @tymofij about the test, and it should cover deletes too.  Also I don't think index_queryset is needed because it does the same as the default according to haystack docs.

jimabramson:
Agree with @tymofij about the test, and it should cover deletes too.  Also I don't think index_queryset is needed because it does the same as the default according to haystack docs.

olmar:
@tymofij  @jimabramson tests added.
Deletion from index is possible only with ``--remove``.  And it seems ``-start/end`` does not play any role when using them with remove flag. See tests for details and this line from django-haystack: https://github.com/toastdriven/django-haystack/blob/master/haystack/management/commands/update_index.py#L244

notesapi/v1/tests/test_update_index.py
```diff
@@ -0,0 +1,79 @@
...
+        url = reverse('api:v1:annotations_detail', kwargs={'annotation_id': first_note['id']})
+        response = self.client.delete(url, self.headers)
+
+        # Delete second note.
+        second_start = datetime.datetime.now()
+        url = reverse('api:v1:annotations_detail', kwargs={'annotation_id': second_note['id']})
+        response = self.client.delete(url, self.headers)
+        second_end = datetime.datetime.now() + datetime.timedelta(minutes=10)
+
+        # Try to update when second note was deleted only.
```
tymofij: when only second note was deleted

olmar: > when only second note was deleted
Fixed.

tymofij:
:+1:

PR #19, opened 2015-01-29
Change config filename ref to match standard

maxrothman:
The DevOps team recently standardized on underscores, rather than dashes, as filename separators.

### DIFF
--- a/notesserver/settings/yaml_config.py
+++ b/notesserver/settings/yaml_config.py
@@ -20,7 +20,7 @@ if not EDXNOTES_CONFIG_ROOT:
 
 CONFIG_ROOT = path(EDXNOTES_CONFIG_ROOT)
 
-with open(CONFIG_ROOT / "edx-notes-api.yml") as yaml_file:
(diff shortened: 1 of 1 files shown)


singingwolfboy:
Where is the companion change to this change? If we're changing from reading `edx-notes-api.yml` to reading `edx_notes_api.yml`, I just want to be sure that something else is similarly changing from writing `edx-notes-api.yml` to writing `edx_notes_api.yml` (or even better, writing both).

maxrothman:
@singingwolfboy Yep, that commit is on its way to https://github.com/edx/configuration/pull/1802. It hasn't gotten pushed yet.

feanil:
@maxrothman looks like tests are failing with this change.

maxrothman:
@singingwolfboy companion change has been pushed. See [this](https://github.com/edx/configuration/blob/max/dash2underscore/playbooks/roles/edx_service/tasks/main.yml#L66) and [this](https://github.com/edx/configuration/blob/max/dash2underscore/playbooks/roles/edx_notes_api/defaults/main.yml#L87).

singingwolfboy:
:+1:

PR #22, opened 2015-04-25
Add tags to model.

cahrens:
https://openedx.atlassian.net/browse/TNL-1923

Note that there is no API to get the set of tags across all notes. I think that we will build up the data structure needed for the page that sorts notes by tags on the client side (since it already gets the complete list of tags).

@dianakhuang and @dan-f Please review

### DIFF
--- a/notesapi/v1/migrations/0001_initial.py
+++ b/notesapi/v1/migrations/0001_initial.py
@@ -1,0 +1,34 @@
+# -*- coding: utf-8 -*-
+""" Initial migration file for creating Note model """
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+    """ Initial migration file for creating Note model """
+
+    dependencies = [
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='Note',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('user_id', models.CharField(
+                    help_text=b'Anonymized user id, not course specific', max_length=255, db_index=True
+                )),
+                ('course_id', models.CharField(max_length=255, db_index=True)),
+                ('usage_id', models.CharField(help_text=b'ID of XBlock where the text comes from', max_length=255)),
+                ('quote', models.TextField(default=b'')),
+                ('text', models.TextField(default=b'', help_text=b"Student's thoughts on the quote", blank=True)),
+                ('ranges', models.TextField(help_text=b'JSON, describes position of quote in the source text')),
+                ('created', models.DateTimeField(auto_now_add=True)),
+                ('updated', models.DateTimeField(auto_now=True)),
+            ],
+            options={
+            },
+            bases=(models.Model,),
+        ),
+    ]
--- a/notesapi/v1/models.py
+++ b/notesapi/v1/models.py
@@ -15,6 +15,7 @@ class Note(models.Model):
     ranges = models.TextField(help_text="JSON, describes position of quote in the source text")
     created = models.DateTimeField(auto_now_add=True)
     updated = models.DateTimeField(auto_now=True)
+    tags = models.TextField(help_text="JSON, list of comma-separated tags", default="[]")
 
     @classmethod
     def create(cls, note_dict):
@@ -34,6 +35,7 @@ class Note(models.Model):
 
         note_dict['ranges'] = json.dumps(ranges)
         note_dict['user_id'] = note_dict.pop('user', None)
+        note_dict['tags'] = json.dumps(note_dict.get('tags', list()))
 
         return cls(**note_dict)
 
@@ -43,7 +45,6 @@ class Note(models.Model):
         """
         created = self.created.isoformat() if self.created else None
         updated = self.updated.isoformat() if self.updated else None
-
         return {
             'id': str(self.pk),
             'user': self.user_id,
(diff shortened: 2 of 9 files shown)


notesapi/v1/migrations/0001_initial.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
```
dan-f: I'm surprised we didn't have an initial migration already!

notesapi/v1/models.py
```diff
@@ -35,6 +36,9 @@ def create(cls, note_dict):
         note_dict['ranges'] = json.dumps(ranges)
         note_dict['user_id'] = note_dict.pop('user', None)
 
+        tags = note_dict.get('tags', list())
```
dan-f: Nit: I think the empty list literal is a little more immediately recognizable.

notesapi/v1/models.py
```diff
@@ -35,6 +36,9 @@ def create(cls, note_dict):
...
+        note_dict['tags'] = json.dumps(tags)
```
dan-f: Nit: why not inline this?

notesapi/v1/models.py
```diff
@@ -54,4 +57,5 @@ def as_dict(self):
             'ranges': json.loads(self.ranges),
             'created': created,
             'updated': updated,
+            'tags': json.loads(self.tags) if self.tags else []
```
dan-f: Why is this `if` necessary? json.loads('[]') works fine. Are you worried about null values for self.tags?

notesapi/v1/tests/test_models.py
```diff
@@ -47,3 +48,12 @@ def test_must_have_fields_create(self):
...
                 note.full_clean()
+
+    def test_default_tags_value(self):
+        """ Test that a note without explicit tags has an empty list for the tags value """
+        payload = self.note_dict.copy()
+        payload.pop("tags")
+        note = Note.create(payload)
+        note.full_clean()
+        note.save()
+        self.assertEqual([], note.as_dict()["tags"])
```
dan-f: Must: I think you should also validate that `note.tags` is what you expect it to be; not just `note.as_dict()["tags"]`.

notesapi/v1/migrations/0001_initial.py
cahrens: Don't get me talking about it... Friday was a frustrating day. This project is using Django 1.7, which handles migrations differently.

notesapi/v1/tests/test_models.py
```diff
@@ -20,6 +20,7 @@ def setUp(self):
                     "endOffset": 10,
                 }
             ],
+            "tags": [u"apple", u"pear"],
```
dan-f: Must: why add a non-empty "tags" list here if the only test in this file which exercises tags removes it?

notesapi/v1/models.py
cahrens: Yes, I was. I wanted to make sure the API is robust to whether or not notes are present (whether or not the plugin is being used).

notesapi/v1/tests/test_models.py
cahrens: There is another test point that creates a note and verifies all the fields work. I didn't have to change that test method at all-- it just worked.

notesapi/v1/tests/test_views.py
```diff
@@ -278,17 +289,34 @@ def test_update(self):
...
+            payload.update(updated_fields)
+            payload.update(self.headers)
+            url = reverse('api:v1:annotations_detail', kwargs={'annotation_id': data['id']})
+            response = self.client.put(url, payload, format='json')
+            self.assertEqual(response.status_code, status.HTTP_200_OK)
+            return response, self._get_annotation(data['id'])
 
-        annotation = self._get_annotation(data['id'])
+        data = self._create_annotation(text=u"Foo")
+        response, annotation = do_update(data, {'id': data['id'], 'text': 'Bar', 'tags': ["new", "tags"]})
```
dan-f: I think that these tags test cases should either be split up into unique test methods, or ddt'd together into one tags-related test method. Either way, I think we should have methods dedicated to testing different aspects of an update, rather than one catch-all method.

notesapi/v1/models.py
dan-f: But doesn't `default="[]"` give us this for free?

notesapi/v1/tests/test_models.py
dan-f: I see. Thanks!

notesapi/v1/models.py
cahrens: Yeah, it probably does. I will try removing it.

notesapi/v1/tests/test_models.py
cahrens: This feels like a bit of an implementation detail that is not necessary to test, but since it is easy enough, I will add it.

cahrens:
@dan-f I responded to your feedback

dan-f:
:+1: looks great!

notesapi/v1/models.py
```diff
@@ -27,13 +28,14 @@ def create(cls, note_dict):
         if len(note_dict) == 0:
             raise ValidationError('Note must have a body.')
 
-        ranges = note_dict.get('ranges', list())
+        ranges = note_dict.get('ranges', [])
```
dianakhuang: Why did you switch this to brackets?

cahrens: Because @dan-f asked me to. Really I have absolutely no preference. Happy to change it back. I think then I would also use list() for tags.

dianakhuang:
That was my only comment. Once that's fixed up, :+1:

PR #23, opened 2015-04-30
SearchIndex must know about tags to return them.

cahrens:
I was not aware of the SearchIndex class when I added tags to the model.

### DIFF
--- a/notesapi/v1/tests/test_views.py
+++ b/notesapi/v1/tests/test_views.py
@@ -417,8 +417,8 @@ class AnnotationSearchViewTests(BaseAnnotationViewTests):
         """
         Tests for search method.
         """
-        self._create_annotation(text=u'First one')
-        self._create_annotation(text=u'Second note')
+        self._create_annotation(text=u'First one', tags=[])
+        self._create_annotation(text=u'Second note', tags=[u'tag1', u'tag2'])
         note = self._create_annotation(text=u'Third note')
         del note['created']
         del note['updated']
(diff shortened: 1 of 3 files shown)


cahrens:
@andy-armstrong and @dan-f Please review. This is a follow-up PR from #22.

(fyi @dianakhuang, who reviewed the first PR)

andy-armstrong:
:+1:

notesapi/v1/tests/test_views.py
```diff
@@ -430,10 +430,16 @@ def test_search(self):
         self.assertEqual(last_note, note)
         self.assertEqual(results['total'], 3)
 
-        results = self._get_search_results(text="Second")
-        self.assertEqual(results['total'], 1)
-        self.assertEqual(len(results['rows']), 1)
-        self.assertEqual(results['rows'][0]['text'], 'Second note')
+        def test_search(searchText, expectedText, expectedTags):
```
dan-f: I think this should have a different name from the encapsulating method to avoid confusion.

dan-f:
@cahrens - does this PR simply make it so that we return tags in search results, or does it also allow for us to search by tag?

cahrens:
@dan-f It only returns the tags as part of the search result model information. It does not allow searching by tags.

notesapi/v1/tests/test_views.py
cahrens: You may have a point there...

dan-f:
Cool, then :+1: from me, once you rename that function.

PR #24, opened 2015-05-14
Search and highlight tags in student notes TNL-1927

benmcmorran:
Changes the search query so that notes are a hit if there is a match in the text or tags. Highlights matching tags. Changes the ordering of the search results when running on ElasticSearch from most recently updated to most relevant first.

Please review @cahrens @dan-f

### DIFF
--- a/notesapi/v1/search_indexes.py
+++ b/notesapi/v1/search_indexes.py
@@ -12,6 +12,7 @@ class NoteIndex(indexes.SearchIndex, indexes.Indexable):
     created = indexes.DateTimeField(model_attr='created')
     updated = indexes.DateTimeField(model_attr='updated')
     tags = indexes.CharField(model_attr='tags')
+    data = indexes.CharField(use_template=True)
 
     def get_model(self):
         return Note
--- a/notesapi/v1/tests/test_views.py
+++ b/notesapi/v1/tests/test_views.py
@@ -10,6 +10,7 @@ from django.core.management import call_command
 from django.core.urlresolvers import reverse
 from django.conf import settings
 from django.http import QueryDict
+from django.test.utils import override_settings
 
 from rest_framework import status
 from rest_framework.test import APITestCase
@@ -481,11 +482,13 @@ class AnnotationSearchViewTests(BaseAnnotationViewTests):
         results = self._get_search_results(text="first", highlight=True, highlight_tag='tag', highlight_class='klass')
         self.assertEqual(results['rows'][0]['text'], '<tag class="klass">First</tag> note')
 
-    def test_search_ordering(self):
-        """
-        Tests ordering of search results.
 
-        Sorting is by descending order by updated field (most recent first).
+    @override_settings(ES_DISABLED=True)
+    def test_search_ordering_in_db(self):
+        """
+        Tests ordering of search results from MySQL.
+
+        MySQL sorting is by descending order by updated field (most recent first).
         """
         self._create_annotation(text=u'First one')
         note = self._create_annotation(text=u'Second note')
@@ -501,6 +504,24 @@ class AnnotationSearchViewTests(BaseAnnotationViewTests):
         self.assertEqual(results['rows'][0]['text'], 'Updated Second Note')
         self.assertEqual(results['rows'][1]['text'], 'Third note')
         self.assertEqual(results['rows'][2]['text'], 'First one')
+
+    @unittest.skipIf(settings.ES_DISABLED, "MySQL does not do relevance ordering")
+    def test_search_ordering_in_es(self):
+        """
+        Tests order of search results from ElasticSearch.
+
+        ElasticSearch sorting is based on the computed relevance of each hit.
+        """
+        self._create_annotation(text=u'fox of the foxes')
+        self._create_annotation(text=u'a very long entry that contains the word fox')
+        self._create_annotation(text=u'the lead fox')
+        self._create_annotation(text=u'does not mention the word')
+
+        results = self._get_search_results(text='fox')
+        self.assertEqual(results['total'], 3)
+        self.assertEqual(results['rows'][0]['text'], 'fox of the foxes')
+        self.assertEqual(results['rows'][1]['text'], 'the lead fox')
+        self.assertEqual(results['rows'][2]['text'], 'a very long entry that contains the word fox')
 
     @unittest.skipIf(settings.ES_DISABLED, "Unicode support in MySQL is limited")
     def test_search_unicode(self):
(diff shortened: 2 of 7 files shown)


notesserver/highlight.py
```diff
@@ -39,13 +39,28 @@ def build_search_kwargs(self, *args, **kwargs):
...
-                }
+                    'tags': {'store': 'yes'},
+                },
             }
             if isinstance(highlight, dict):
                 highlight_options.update(highlight)
             res['highlight'] = highlight_options
         return res
 
+    def _process_results(
```
benmcmorran: @jimabramson I was struggling with the correct way to return the highlighted tags from haystack. It looks like haystack can only return the highlight from the content field (`text` in this case) using the ElasticsearchSearchBackend: https://github.com/django-haystack/django-haystack/blob/be67c37e3d23eed96d675e0c56e85eb6de0d484b/haystack/backends/elasticsearch_backend.py#L614. However, we need the highlight information for both the text and the tags. The code below is a bit of a hack, and I'm open to cleaner ways of returning this information.

notesserver/highlight.py
```diff
@@ -39,13 +39,28 @@ def build_search_kwargs(self, *args, **kwargs):
...
+            raw_results, highlight, result_class, distance_point, geo_sort
+        )
+
+        for i, raw_result in enumerate(raw_results.get('hits', {}).get('hits', [])):
+            if 'highlight' in raw_result:
+                result['results'][i].highlighted_tags = raw_result['highlight'].get('tags', '')
+
+        return result
+
+
```
cahrens: Extra newline

notesapi/v1/views.py
```diff
@@ -64,7 +71,7 @@ def get_from_es(self, *args, **kwargs):  # pylint: disable=unused-argument
             opts = {
                 'pre_tags': ['<{tag}{klass_str}>'.format(
                     tag=tag,
-                    klass_str=' class="{}"'.format(klass) if klass else ''
+                    klass_str=' class=\\"{}\\"'.format(klass) if klass else ''
```
cahrens: Can you explain this change?

benmcmorran: The tags are stored in the database as a single JSON string array in string from. For example, `tags` might be `["foo", "bar"]` as a string. If we insert the highlight tag around `foo` without a `\` then we could get `["<span class="highlight">foo</span>", "bar"]`. This is no longer valid JSON. Instead the quotes in the injected tag are escaped, so we get `["<span class \"=highlight\">foo</span>", "bar"]` which can be deserialized successfully.

cahrens:
:+1: I am interested though in what @jimabramson has to say about "the hack".

notesapi/v1/tests/test_views.py
```diff
@@ -502,6 +505,24 @@ def test_search_ordering(self):
...
+        self._create_annotation(text=u'fox of the foxes')
+        self._create_annotation(text=u'the lead fox')
+        self._create_annotation(text=u'a very long entry that contains the word fox')
+        self._create_annotation(text=u'does not mention the word')
+
+        results = self._get_search_results(text='fox')
+        self.assertEqual(results['total'], 3)
+        self.assertEqual(results['rows'][0]['text'], 'fox of the foxes')
+        self.assertEqual(results['rows'][1]['text'], 'the lead fox')
+        self.assertEqual(results['rows'][2]['text'], 'a very long entry that contains the word fox')
```
jimabramson: are you sure they aren't being returned in the order they were created, rather than by relevance?

benmcmorran: Good point. I'll flip the order of creation for the last two.

notesserver/highlight.py
jimabramson: @benmcmorran did you consider moving the tag highlighting closer to the client, rather than making it look like haystack/ES is doing it?  I've always considered the insertion of this kind of markup from the server a little ugly and anti-encapsulation, but in the case of highlighting match text within a larger body I can understand the value.  When you're rendering tags tho, isn't it just a simple 'if' statement to style them differently?

benmcmorran: I'll try that. I wasn't sure about the performance impact of doing highlighting outside of ES, but @dan-f is working on measuring that.

benmcmorran: @jimabramson Wouldn't moving highlighting logic into the API handlers require them to handle many of the concerns that ES currently deals with, like tokenizing and stemming of searches? The user could potentially search for both a tag and text in the same query. This approach could end up being messier than the current one.

jimabramson: to clarify, i only meant to propose highlighting tags on the client, not text in the body of the document.  this is based on some assumptions i'm making, for example that you're analyzing / matching tags as keywords, rather than doing any tokenization / stemming there.  assuming that, frontend code could check tags as they're being rendered and do a simple equality check to decide whether they should be styled specially (highlighted) or not.  another assumption i'm making is that the frontend is built in such a way that this is feasible (e.g. you have easy access to the text that was searched at the moment you are rendering the tags in the results).

jimabramson:
@benmcmorran i don't know how well this relates to the requirements, and i don't know if it's supported in haystack, but you might consider taking a quick look at elastic's [faceting docs](http://www.elastic.co/guide/en/elasticsearch/reference/0.90/search-facets.html) and see if using any of those features seem relevant / worth exploring.

dan-f:
Hey @benmcmorran when you're all done with code review, be sure to rebase the ticket number into your commits / add it to this PR.

explorerleslie:
@cahrens :+1: this looks fantastic!

benmcmorran:
@cahrens @dan-f I pushed a few minor changes to fix an issue I found where the search had to completely match in either tags or text. Now ElasticSearch builds the concatenation of tags and text into the index and searches in this field to support queries like "tag1 tag2 text"

notesapi/v1/search_indexes.py
```diff
@@ -12,6 +12,7 @@ class NoteIndex(indexes.SearchIndex, indexes.Indexable):
     created = indexes.DateTimeField(model_attr='created')
     updated = indexes.DateTimeField(model_attr='updated')
     tags = indexes.CharField(model_attr='tags')
+    data = indexes.CharField(use_template=True)
```
cahrens: Does this mean we have to recreate the search index (like I had to when I added the tags field right above)?

benmcmorran: Yes, the index does need to be recreated. I'm not sure about the performance impact. @dan-f any thoughts?

cahrens:
I updated the sandbox for this commit and rebuilt the search index.

http://tags.m.sandbox.edx.org/courses/edX/DemoX.1/2015/edxnotes/

:+1: It seems like we have to do this to get search to work properly, and Dan can look at the impact as part of his performance work.

We will want to update https://openedx.atlassian.net/wiki/display/TNL/Student+Notes+Tagging to make it clear that the index needs to rebuilt for this story as well.

notesapi/v1/search_indexes.py
dan-f: My only thought is that I'd expect this to increase memory/disk usage, since we've not just added text and tags to the index but the composite text+tags field as well. Should we consider removing the individual text/tags fields from the index? @jimabramson

notesapi/v1/tests/test_views.py
dan-f: I like this example :)

notesapi/v1/tests/test_views.py
```diff
@@ -546,6 +567,93 @@ def test_search_course(self):
...
+        self._create_annotation(text=u'Another one', tags=[u'bar'])
+        self._create_annotation(text=u'A third note', tags=[u'bar', u'baz'])
+        self._create_annotation(text=u'One final note', tags=[])
+
+        results = self._get_search_results(text='Foo')
+        self.assertEqual(results['total'], 1)
+        self.assertEqual(results['rows'][0]['text'], 'First note')
+
+        results = self._get_search_results(text='bar')
+        self.assertEqual(results['total'], 3)
```
dan-f: I think you should assert that each of the notes are what you expect them to be, since in this case you get multiple results; so far we have only inspected the notes when we get one result. You could write a helper method and then call it as well from `test_search_ordering_in_es()` and `_search_tag_or_text()`.

notesapi/v1/views.py
```diff
@@ -45,7 +48,7 @@ def get_from_db(self, *args, **kwargs):  # pylint: disable=unused-argument
             query = query.filter(user_id=params['user'])
 
         if 'text' in params:
-            query = query.filter(text__icontains=params['text'])
+            query = query.filter(Q(text__icontains=params['text']) | Q(tags__icontains=params['text']))
```
dan-f: Shouldn't we now be using the `data` field you added to the index?

dan-f: Oops just saw that this is the database search.

notesserver/highlight.py
```diff
@@ -39,13 +39,27 @@ def build_search_kwargs(self, *args, **kwargs):
...
+                },
             }
             if isinstance(highlight, dict):
                 highlight_options.update(highlight)
             res['highlight'] = highlight_options
         return res
 
+    def _process_results(
+            self, raw_results, highlight=False, result_class=None, distance_point=None, geo_sort=False
+    ):
```
dan-f: Don't forget to add a docstring to this method.

templates/search/indexes/v1/note_data.txt
```diff
@@ -0,0 +1,2 @@
+{{ object.text }}
+{{ object.tags }}
```
dan-f: Nit: missing trailing newline.

dan-f:
@benmcmorran: Looks really great to me! :+1:, except don't actually merge until we do performance testing.

notesapi/v1/tests/test_views.py
benmcmorran: I've added a helper to check the text of the results without respect to order. My thinking here is that order should have already been tested by the ordering tests. The other tests should succeed regardless of which backend is doing the searching.

dan-f: Thanks.

dan-f:
P.s. don't forget to add yourself to AUTHORS.

PR #25, opened 2015-06-04
Add management command to create notes in bulk

dan-f:
This script affords the bulk creation of notes. It takes as input:
* number of notes
* number of notes that should be attributed to the same user
* a set of `course_id`s to randomly be attributed to notes

It tries to have reasonable heuristics/weighted probabilities for note content length.

@jibsheet @feanil please review


notesapi/v1/management/commands/bulk_create_notes.py
```diff
@@ -0,0 +1,69 @@
...
+
+class Command(BaseCommand):
+    args = '<total_notes> <notes_per_user> <course_id course_id ...>'
+    help = 'Add N random notes to the database'
+
+    def handle(self, *args, **options):
+        if len(args) < 3:
+            raise CommandError("generate_random_notes takes the following arguments: " + self.args)
+
+        total_notes = int(args[0])
```
feanil: Please use the option parser to do validation on the command input. https://docs.djangoproject.com/en/1.4/howto/custom-management-commands/

dan-f: I've seen the wiki page. At the time it seemed like all inputs were arguments more than options. I'll refactor `notes_per_user` and `course_ids`  to be options with sane defaultssss.

dan-f:
@feanil - I addressed your comment regarding optparse.

feanil:
:+1:

PR #31, opened 2015-12-01
Paginate API

muhammad-ammar:
## [TNL-2453](https://openedx.atlassian.net/browse/TNL-2453)

This adds the pagination to  the edx-notes apis via DRF.

### Testing
- [x] Unit

### Reviewers
- [x] Code review: @symbolist 
- [x] Code review: @muzaffaryousaf

### Post-review
- [x] Squash commits

### DIFF
--- a/notesapi/v1/paginators.py
+++ b/notesapi/v1/paginators.py
@@ -1,0 +1,27 @@
+"""
+Paginator for Notes.
+"""
+
+from rest_framework import pagination
+from rest_framework.response import Response
+
+
+class NotesPaginator(pagination.PageNumberPagination):
+    """
+    Student Notes Paginator.
+    """
+    page_size = 10
+    page_size_query_param = "page_size"
+
+    def get_paginated_response(self, data):
+        """
+        Annotate the response with pagination information.
+        """
+        return Response({
+            'current': self.page.number,
+            'next': self.get_next_link(),
+            'previous': self.get_previous_link(),
+            'count': self.page.paginator.count,
+            'num_pages': self.page.paginator.num_pages,
+            'results': data
+        })
--- a/notesapi/v1/paginators.py
+++ b/notesapi/v1/paginators.py
@@ -22,7 +22,7 @@ class NotesPaginator(pagination.PageNumberPagination):
             'current_page': self.page.number,
             'next': self.get_next_link(),
             'previous': self.get_previous_link(),
-            'count': self.page.paginator.count,
+            'total': self.page.paginator.count,
             'num_pages': self.page.paginator.num_pages,
-            'results': data
+            'rows': data
         })
--- a/notesapi/v1/paginators.py
+++ b/notesapi/v1/paginators.py
@@ -18,7 +18,8 @@ class NotesPaginator(pagination.PageNumberPagination):
         Annotate the response with pagination information.
         """
         return Response({
-            'current': self.page.number,
+            'start': (self.page.number - 1) * self.get_page_size(self.request),
+            'current_page': self.page.number,
             'next': self.get_next_link(),
             'previous': self.get_previous_link(),
             'count': self.page.paginator.count,
--- a/notesapi/v1/paginators.py
+++ b/notesapi/v1/paginators.py
@@ -5,12 +5,14 @@ Paginator for Notes.
 from rest_framework import pagination
 from rest_framework.response import Response
 
+from django.conf import settings
+
 
 class NotesPaginator(pagination.PageNumberPagination):
     """
     Student Notes Paginator.
     """
-    page_size = 10
+    page_size = settings.DEFAULT_NOTES_PAGE_SIZE
     page_size_query_param = "page_size"
 
     def get_paginated_response(self, data):
--- a/notesapi/v1/serializers.py
+++ b/notesapi/v1/serializers.py
@@ -56,7 +56,7 @@ class NotesElasticSearchSerializer(serializers.Serializer):  # pylint: disable=a
         Return note text.
         """
         if note.highlighted:
-            return note.highlighted[0].decode('unicode_escape')
+            return note.highlighted[0]
         return note.text
 
     def get_ranges(self, note):
--- a/notesapi/v1/serializers.py
+++ b/notesapi/v1/serializers.py
@@ -1,0 +1,74 @@
+"""
+Serializers for Notes API.
+"""
+
+import json
+from rest_framework import serializers
+from notesapi.v1.models import Note
+
+
+class NoteSerializer(serializers.ModelSerializer):
+    """
+    Student Notes Model Serializer.
+    """
+    class Meta(object):
+        """
+        Model Serializer Meta Class.
+        """
+        model = Note
+        exclude = ('user_id',)
+
+    id = serializers.CharField(source='pk')
+    user = serializers.CharField(source='user_id')
+    ranges = serializers.SerializerMethodField()
+    tags = serializers.SerializerMethodField()
+
+    def get_ranges(self, note):
+        """
+        Return note ranges.
+        """
+        return json.loads(note.ranges)
+
+    def get_tags(self, note):
+        """
+        Return note tags.
+        """
+        return json.loads(note.tags)
+
+
+class NotesElasticSearchSerializer(serializers.Serializer):  # pylint: disable=abstract-method
+    """
+    Student Notes Elastic Search Serializer.
+    """
+    id = serializers.CharField(source='pk')
+    user = serializers.CharField()
+    created = serializers.DateTimeField()
+    updated = serializers.DateTimeField()
+    quote = serializers.CharField()
+    course_id = serializers.CharField()
+    usage_id = serializers.CharField()
+    text = serializers.SerializerMethodField()
+    ranges = serializers.SerializerMethodField()
+    tags = serializers.SerializerMethodField()
+
+    def get_text(self, note):
+        """
+        Return note text.
+        """
+        if note.highlighted:
+            return note.highlighted[0].decode('unicode_escape')
+        return note.text
+
+    def get_ranges(self, note):
+        """
+        Return note ranges.
+        """
+        return json.loads(note.ranges)
+
+    def get_tags(self, note):
+        """
+        Return note tags.
+        """
+        if note.highlighted_tags:
+            return json.loads(note.highlighted_tags[0])
+        return json.loads(note.tags) if note.tags else []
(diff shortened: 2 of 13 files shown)


muhammad-ammar:
@muzaffaryousaf @symbolist This is ready for review. PTAL. Thanks!

notesapi/utils.py
```diff
@@ -0,0 +1,124 @@
...
+            'current': self.page.number,
+            'next': self.get_next_link(),
+            'previous': self.get_previous_link(),
+            'count': self.page.paginator.count,
+            'num_pages': self.page.paginator.num_pages,
+            'results': data
+        })
+
+
+class NotesSerializer(serializers.ModelSerializer):
```
muzaffaryousaf: Move all serializers to the serializers.py.

notesapi/utils.py
```diff
@@ -0,0 +1,124 @@
...
+    Student Notes Model Serializer.
+    """
+    class Meta(object):
+        """
+        Model Serializer Meta Class.
+        """
+        model = Note
+        exclude = ('user_id',)
+
+    id = serializers.SerializerMethodField()
```
muzaffaryousaf: Don't we have this field in model. Can we use that field as it, if not, we can make field like this --> `serializers.Field(source='SourceFiledName')`

notesapi/utils.py
```diff
@@ -0,0 +1,124 @@
...
+    user = serializers.SerializerMethodField()
```
muzaffaryousaf: We can use user_id instead of this custom filed, just remove exclude statement above.

notesapi/utils.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+        Return string representation of note primary key.
+        """
+        return str(note.pk)
+
+    def get_user(self, note):
+        """
+        Return string representation of note user_id.
+        """
+        return str(note.user_id)
```
muzaffaryousaf: user_id is anonymous user id not the actual id. I hope you already knew it.

notesapi/v1/views.py
```diff
@@ -77,26 +82,17 @@ def get_from_es(self, *args, **kwargs):  # pylint: disable=unused-argument
...
-
-        return results
+        page = self.paginate_queryset(query)
+        serializer = NotesElasticSearchSerializer(page, many=True)
+        response = self.get_paginated_response(serializer.data)
+        return response
 
 
-class AnnotationListView(APIView):
+class AnnotationListView(GenericAPIView):
```
muzaffaryousaf: Can we use [ListCreateAPIView](http://www.django-rest-framework.org/api-guide/generic-views/#listcreateapiview) which extends GenericAPIView. We have to convert post method to use serializers for consistency.

notesapi/v1/views.py
```diff
@@ -77,26 +82,17 @@ def get_from_es(self, *args, **kwargs):  # pylint: disable=unused-argument
...
     """
     List all annotations or create.
```
muzaffaryousaf: It would be nice if we can add proper doc strings for these views, which we will be using for documentation. Here is an example https://github.com/edx/edx-platform/pull/8082/files#diff-f0ef24d2da6e0cc7d7c5ce4d53fdda70R71

notesapi/utils.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def get_ranges(self, note):
+        """
+        Return note ranges.
+        """
+        return json.loads(note.ranges)
```
muzaffaryousaf: Can we add properties to the Note model and remove these methods along with other usages where we dump/load these fields.

notesapi/utils.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def get_tags(self, note):
```
muzaffaryousaf: Same as above.

notesapi/utils.py
```diff
@@ -0,0 +1,124 @@
+"""
+Paginator and Serializers.
+"""
+
+import json
+
+from rest_framework import pagination
+from rest_framework import serializers
```
muzaffaryousaf: Can we make it single import.

notesapi/v1/views.py
```diff
@@ -50,7 +52,10 @@ def get_from_db(self, *args, **kwargs):  # pylint: disable=unused-argument
         if 'text' in params:
             query = query.filter(Q(text__icontains=params['text']) | Q(tags__icontains=params['text']))
 
-        return [note.as_dict() for note in query]
```
muzaffaryousaf: Can we remove `as_dict()` method, as we have `NoteSerializer` now, also, we can replace it with NoteSerializers in DetailView too.

notesapi/utils.py
muzaffaryousaf: NIT: `NoteSerializer`.

notesapi/v1/tests/test_views.py
```diff
@@ -98,7 +98,59 @@ def _get_search_results(self, **kwargs):
         result = self.client.get(url, data=data)
         return result.data
 
+    def verify_pagination_info(
+            self, response, total_annotations, num_pages, annotations_per_page, current_page, previous_page, next_page
```
muzaffaryousaf: NIT: can we list params vertically ?

notesapi/v1/tests/test_views.py
```diff
@@ -271,6 +325,74 @@ def test_read_all_no_query_param(self):
...
+            previous_page=previous_page,
+            next_page=next_page
+        )
+
+    def test_pagination_single_page(self):
+        """
+        Verify that pagination info is correct when we have a single page of data.
+        """
+        total_annotations = 6
+        for aid in range(total_annotations):
```
muzaffaryousaf: `aid` stands for ?

notesapi/v1/tests/test_views.py
```diff
@@ -271,6 +325,74 @@ def test_read_all_no_query_param(self):
...
+        response = self.get_annotations(query_parameters={'page_size': page_size})
+        self.verify_pagination_info(
+            response,
+            total_annotations=total_annotations,
+            num_pages=num_pages,
+            annotations_per_page=annotations_per_page,
+            current_page=1,
+            previous_page=None,
+            next_page=next_page
+        )
```
muzaffaryousaf: Can we create a helper method which has all these statements to avoid the duplications ? We can call that method from these tests.

notesapi/v1/tests/test_views.py
```diff
@@ -651,18 +774,86 @@ def test_search_highlight_tag(self):
...
+        response = self._get_search_results(page_size=page_size)
+        self.verify_pagination_info(
+            response,
+            total_annotations=total_annotations,
+            num_pages=num_pages,
+            annotations_per_page=annotations_per_page,
+            current_page=1,
+            previous_page=None,
+            next_page=next_page
+        )
```
muzaffaryousaf: We can have a helper method for these tests too.

muzaffaryousaf:
@muhammad-ammar I'm done with first pass and left some comments. Let me know when you are done.

muzaffaryousaf:
It would be nice if we try to follow this template for PRs https://openedx.atlassian.net/wiki/display/TNL/TNL+Pull+Request+Template

notesapi/utils.py
muhammad-ammar: I would prefer to remain these as is.

muhammad-ammar:
@muzaffaryousaf I have addressed your feedback. Please continue the review. Thanks!

notesapi/v1/serializers.py
```diff
@@ -0,0 +1,74 @@
...
+class NoteSerializer(serializers.ModelSerializer):
+    """
+    Student Notes Model Serializer.
+    """
+    class Meta(object):
+        """
+        Model Serializer Meta Class.
+        """
+        model = Note
+        exclude = ('user_id',)
```
symbolist: It nicer to have a `fields` attribute rather than an `exclude` attribute because it makes it clear what will be added.

notesapi/v1/serializers.py
```diff
@@ -0,0 +1,74 @@
...
+        """
+        return json.loads(note.tags)
+
+
+class NotesElasticSearchSerializer(serializers.Serializer):  # pylint: disable=abstract-method
+    """
+    Student Notes Elastic Search Serializer.
+    """
+    id = serializers.CharField(source='pk')
+    user = serializers.CharField(source='user_id')
```
symbolist: I note in the logic in the view, that this replaces, `user` is not being picked from `user_id`. Can you check if we need to do this or will the elastic search responses only have the `user` field?

symbolist:
@muhammad-ammar This is looking great! I only have one question.

muhammad-ammar:
@muzaffaryousaf @symbolist Addressed your feedback. Please continue the review. Thanks!

symbolist:
:+1:

notesapi/v1/paginators.py
```diff
@@ -0,0 +1,27 @@
+"""
+Paginator and Serializers.
```
muzaffaryousaf: Remove serializers.

notesapi/v1/views.py
```diff
@@ -131,15 +294,78 @@ def post(self, *args, **kwargs):  # pylint: disable=unused-argument
...
+
+            * tags: List. Comma separated tags.
+
+            * created: DateTime. Creation datetime of annotation.
+
+            * updated: DateTime. When was the last time annotation was updated.
+
+        **Response Values for PUT**
+
+            * same as GET
```
muzaffaryousaf: Don't we send dict that we will be updating.

muzaffaryousaf:
:+1: except 2 NITs.

muzaffaryousaf:
@symbolist Should we wait for https://github.com/edx/edx-notes-api/pull/44 ?

symbolist:
@muzaffaryousaf Lets merge this.

nedbat:
Removing requirements/optional.txt broke installations using the configuration repo.  Was there something else referencing optional.txt?

muzaffaryousaf:
@nedbat We removed the optional.txt from configuration too but looks like we are using old version of it in dogwood. [Here](https://github.com/edx/configuration/pull/2848) is the PR.

For now, i'm putting an empty optional.txt file just to make sure that we don't break with old configuration repo. Does it sounds okay to you?

nedbat:
Thanks, I did find that configuration PR after I put the comment here.  I think the real problem is that Dogwood did not tag this repo, and does not try to install a specific version.  So we are getting master of edx-notes-api.  An empty optional.txt will solve the immediate problem, but I don't know what else will be broken if we use master of edx-notes-api with Dogwood of everything else. 

The right fix is to fix the Dogwood install, which I'll be doing.  Thanks.

nedbat:
Just to be clear: I think I prefer that you *not* add back an empty optional.txt.  At least the way things are now, we have a very clear failure to indicate that the versions are skewed.

PR #37, opened 2016-02-04
fix unicode error when search with highlighting

muhammad-ammar:
## [TNL-4101](https://openedx.atlassian.net/browse/TNL-4101)

During serialization unicode string is using decode method to decode itself using unicode_escape codec. Which is incorrect because decode method is for byte string not for unicode string.

### Testing
- [x] Unit

### Reviewers
- [x] Code review: @symbolist 
- [ ] Code review: @muzaffaryousaf

### Post-review
- [ ] Squash commits

### DIFF
--- a/notesapi/v1/serializers.py
+++ b/notesapi/v1/serializers.py
@@ -56,7 +56,7 @@ class NotesElasticSearchSerializer(serializers.Serializer):  # pylint: disable=a
         Return note text.
         """
         if note.highlighted:
-            return note.highlighted[0].decode('unicode_escape')
+            return note.highlighted[0]
         return note.text
 
     def get_ranges(self, note):
--- a/notesapi/v1/tests/test_views.py
+++ b/notesapi/v1/tests/test_views.py
@@ -738,6 +738,25 @@ class AnnotationSearchViewTests(BaseAnnotationViewTests):
         response = self._get_search_results(text=u"Свят")
         self.assertEqual(response['rows'][0]['text'], u'Веселих свят')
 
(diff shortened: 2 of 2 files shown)


muhammad-ammar:
@symbolist @muzaffaryousaf This is ready for review.

notesapi/v1/serializers.py
```diff
@@ -56,7 +56,7 @@ def get_text(self, note):
         Return note text.
         """
         if note.highlighted:
-            return note.highlighted[0].decode('unicode_escape')
```
symbolist: @muhammad-ammar Do you know why `decode` was here in the first place?

muhammad-ammar: @symbolist unfortunately not.

muzaffaryousaf: Maybe someone from Cambridge can tell this ?

muhammad-ammar: @andy-armstrong Do you remember why we have added `decode` method here?

andy-armstrong: I wasn't involved in this work, so I don't know. It does seem like a strange thing to be doing. @jimabramson do you have any insights?

muhammad-ammar: @jimabramson this `decode` method change was added in https://github.com/edx/edx-notes-api/pull/24

symbolist:
@muhammad-ammar As we discussed, this looks good. Lets go ahead with merging it. :+1:

PR #39, opened 2016-02-08
[TNL-4029] user can't add more than maximum number of allowed notes

ehteshamkafeel:
Throws an error if tried to create more notes per course than defined in settings.
@symbolist @muzaffaryousaf @muhammad-ammar please review


ehteshamkafeel:
Pull request moved [here]( https://github.com/edx/edx-notes-api/pull/40)

PR #40, opened 2016-02-10
[TNL-4029] user can't add more than maximum number of allowed notes

ehteshamkafeel:
Throws an error if tried to create more notes per course than defined in settings.
@symbolist @muzaffaryousaf @muhammad-ammar please review

### DIFF
--- a/notesapi/v1/tests/test_views.py
+++ b/notesapi/v1/tests/test_views.py
@@ -18,6 +18,7 @@ from rest_framework.test import APITestCase
 from .helpers import get_id_token
 
 TEST_USER = "test_user_id"
+TEST_OTHER_USER = "test_other_user_id"
 
 if not settings.ES_DISABLED:
     import haystack
@@ -54,7 +55,7 @@ class BaseAnnotationViewTests(APITestCase):
             "tags": ["pink", "lady"]
         }
 
-    def _create_annotation(self, **kwargs):
+    def _create_annotation(self, expected_status=status.HTTP_201_CREATED, **kwargs):
         """
         Create annotation
         """
@@ -62,7 +63,7 @@ class BaseAnnotationViewTests(APITestCase):
         opts.update(kwargs)
         url = reverse('api:v1:annotations')
         response = self.client.post(url, opts, format='json')
-        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(response.status_code, expected_status)
         return response.data.copy()
 
     def _do_annotation_update(self, data, updated_fields):
@@ -345,6 +346,38 @@ class AnnotationListViewTests(BaseAnnotationViewTests):
         del annotation['updated']
         del annotation['created']
         self.assertEqual(annotation, note)
+
+    @patch('django.conf.settings.MAX_NOTES_PER_COURSE', 5)
+    def test_create_maximum_allowed(self):
+        """
+        Tests if user can create more than maximum allowed notes per course
+        Also test if other user can create notes and Same user can create notes in other course
+        """
+        for i in xrange(5):
+            kwargs = {'text': 'Foo_{}'.format(i)}
+            self._create_annotation(**kwargs)
+
+        # Creating more notes should result in 400 error
+        kwargs = {'text': 'Foo_{}'.format(6)}
+        response = self._create_annotation(expected_status=status.HTTP_400_BAD_REQUEST, **kwargs)
+        self.assertEqual(
+            response['error_msg'],
+            u'You can create up to {0} notes.'
+            u' You must remove some notes before you can add new ones.'.format(settings.MAX_NOTES_PER_COURSE)
+        )
+
+        # if user tries to create note in a different course it should succeed
+        kwargs = {'course_id': 'test-course-id-2'}
+        response = self._create_annotation(**kwargs)
+        self.assertTrue('id' in response)
+
+        # if another user to tries to create note in first course it should succeed
+        token = get_id_token(TEST_OTHER_USER)
+        self.client.credentials(HTTP_X_ANNOTATOR_AUTH_TOKEN=token)
+        self.headers = {'user': TEST_OTHER_USER}
+        kwargs = {'user': TEST_OTHER_USER}
+        response = self._create_annotation(**kwargs)
+        self.assertTrue('id' in response)
 
     def test_read_all_no_annotations(self):
         """
--- a/notesapi/v1/views.py
+++ b/notesapi/v1/views.py
@@ -1,10 +1,12 @@
 import logging
 import json
+import newrelic.agent
 
 from django.conf import settings
 from django.core.urlresolvers import reverse
 from django.core.exceptions import ValidationError
 from django.db.models import Q
+from django.utils.translation import ugettext as _
 
 from rest_framework import status
 from rest_framework.generics import GenericAPIView
@@ -20,6 +22,13 @@ if not settings.ES_DISABLED:
     from notesserver.highlight import SearchQuerySet
 
 log = logging.getLogger(__name__)
+
+
+class AnnotationsLimitReachedError(Exception):
+    """
+    Exception when trying to create more than allowed annotations
+    """
+    pass
 
 
 class AnnotationSearchView(GenericAPIView):
@@ -275,20 +284,41 @@ class AnnotationListView(GenericAPIView):
 
         Returns 400 request if bad payload is sent or it was empty object.
         """
-        if 'id' in self.request.data:
+        if not self.request.data or 'id' in self.request.data:
             return Response(status=status.HTTP_400_BAD_REQUEST)
 
         try:
+            total_notes = Note.objects.filter(
+                    user_id=self.request.data['user'], course_id=self.request.data['course_id']
+            ).count()
+            if total_notes >= settings.MAX_NOTES_PER_COURSE:
+                raise AnnotationsLimitReachedError
+
             note = Note.create(self.request.data)
             note.full_clean()
+
+            # Gather metrics for New Relic so we can slice data in New Relic Insights
+            newrelic.agent.add_custom_parameter('notes.count', total_notes)
         except ValidationError as error:
             log.debug(error, exc_info=True)
             return Response(status=status.HTTP_400_BAD_REQUEST)
+        except AnnotationsLimitReachedError:
+            error_message = _(
+                u'You can create up to {max_num_annotations_per_course} notes.'
+                u' You must remove some notes before you can add new ones.'
+            ).format(max_num_annotations_per_course=settings.MAX_NOTES_PER_COURSE)
+            log.info(
+                u'Attempted to create more than %s annotations',
+                settings.MAX_NOTES_PER_COURSE
+            )
+
+            return Response({
+                'error_msg': error_message
+            }, status=status.HTTP_400_BAD_REQUEST)
 
         note.save()
 
         location = reverse('api:v1:annotations_detail', kwargs={'annotation_id': note.id})
-
         serializer = NoteSerializer(note)
         return Response(serializer.data, status=status.HTTP_201_CREATED, headers={'Location': location})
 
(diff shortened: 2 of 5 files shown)


muzaffaryousaf:
@ehteshamkafeel We need to `newrelic.agent.add_custom_parameter('notes.count', len(notes))` for monitoring purposes. Here are more examples https://github.com/edx/edx-platform/pull/10840/files

notesapi/v1/tests/test_views.py
```diff
@@ -65,6 +66,16 @@ def _create_annotation(self, **kwargs):
...
 
+    def _create_annotation_without_assert(self, **kwargs):
+        """
+        Create annotation but do not check for 201
+        """
+        opts = self.payload.copy()
+        opts.update(kwargs)
+        url = reverse('api:v1:annotations')
+        response = self.client.post(url, opts, format='json')
+        return response
```
muhammad-ammar: @ehteshamkafeel Instead of creating a new method we can use the existing `_create_annotation` and use a kwarg to decide whether we need to check the http status_code.

notesapi/v1/views.py
```diff
@@ -275,20 +283,36 @@ def post(self, *args, **kwargs):  # pylint: disable=unused-argument
...
+            error_message = ugettext_noop(
+                u'You can create up to {max_num_annotations_per_course} notes.'
+                u' You must remove some notes before you can add new ones.'
+            ).format(max_num_annotations_per_course=settings.MAX_ANNOTATIONS_PER_COURSE)
+            log.info(
+                u'Attempted to create more than %s annotations',
+                settings.MAX_ANNOTATIONS_PER_COURSE
+            )
+            return Response({
+                'error_msg': error_message
```
muhammad-ammar: @ehteshamkafeel Shouldn't we need to translate `error_message`?

notesapi/v1/views.py
```diff
@@ -275,20 +283,36 @@ def post(self, *args, **kwargs):  # pylint: disable=unused-argument
...
+            ).count() >= settings.MAX_ANNOTATIONS_PER_COURSE:
+                raise AnnotationsLimitReachedError
+
             note = Note.create(self.request.data)
             note.full_clean()
         except ValidationError as error:
             log.debug(error, exc_info=True)
             return Response(status=status.HTTP_400_BAD_REQUEST)
+        except AnnotationsLimitReachedError:
+            error_message = ugettext_noop(
```
muhammad-ammar: @ehteshamkafeel I think we don't need to use `ugettext_noop` because we not using `error_message` for any intermediate purpose and just returning in response.

notesapi/v1/tests/test_views.py
```diff
@@ -346,6 +357,64 @@ def test_create_multirange(self):
...
+            self.assertEqual(
+                response.data['error_msg'],
+                u'You can create up to {0} notes.'
+                u' You must remove some notes before you can add new ones.'.format(settings.MAX_ANNOTATIONS_PER_COURSE)
+            )
+        else:
+            self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+    @patch('django.conf.settings.MAX_ANNOTATIONS_PER_COURSE', 5)
+    def test_create_maximum_allowed_other(self):
```
muhammad-ammar: @ehteshamkafeel Can we change the method name to something better so that one can get some idea about test by reading the test name.

notesapi/v1/views.py
```diff
@@ -275,20 +283,36 @@ def post(self, *args, **kwargs):  # pylint: disable=unused-argument
 
         Returns 400 request if bad payload is sent or it was empty object.
         """
-        if 'id' in self.request.data:
+        if not self.request.data or 'id' in self.request.data:
```
muhammad-ammar: @ehteshamkafeel Why we need to add this?

muhammad-ammar:
@ehteshamkafeel I am done with 1st pass. Let me know when you are done with changes.

notesapi/v1/tests/test_views.py
ehteshamkafeel: no if I add a third argument I will have to change all calls, if I send the flag via kwargs it will change the payload for post.

notesapi/v1/views.py
ehteshamkafeel: can we discuss that?

notesapi/v1/views.py
ehteshamkafeel: this is needed for the tests or even real cases where request data is empty. As we first check for notes count before creating a note. Validation error raises when we create note.

notesapi/v1/tests/test_views.py
muhammad-ammar:
```python
def _create_annotation(self, check_status=True, **kwargs):
        """
        Create annotation
        """
        opts = self.payload.copy()
        opts.update(kwargs)
        url = reverse('api:v1:annotations')
        response = self.client.post(url, opts, format='json')
        if check_status:
            self.assertEqual(response.status_code, status.HTTP_201_CREATED)
        return response.data.copy()
```

@ehteshamkafeel as per my understanding this should work. you don't need to change all the `_create_annotation` calls

notesapi/v1/views.py
```diff
@@ -15,6 +16,7 @@
 
 from notesapi.v1.models import Note
 from notesapi.v1.serializers import NoteSerializer, NotesElasticSearchSerializer
+import newrelic.agent
```
muzaffaryousaf: Ideally third party imports should be in higher order than our app components.

notesapi/v1/tests/test_views.py
```diff
@@ -346,6 +357,64 @@ def test_create_multirange(self):
...
+                response.data['error_msg'],
+                u'You can create up to {0} notes.'
+                u' You must remove some notes before you can add new ones.'.format(settings.MAX_ANNOTATIONS_PER_COURSE)
+            )
+        else:
+            self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+    @patch('django.conf.settings.MAX_ANNOTATIONS_PER_COURSE', 5)
+    def test_create_maximum_other_user_course(self):
+        # if user tries to create notes in another course it should succeed
```
muzaffaryousaf: Doc string missing.

notesapi/v1/tests/test_views.py
```diff
@@ -346,6 +357,64 @@ def test_create_multirange(self):
...
+
+        # Creating more notes should result in 400 error
+        kwargs = {'text': 'Foo_{}'.format(6)}
+        response = self._create_annotation_without_assert(**kwargs)
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST, 'Creating more than allowed notes')
+        self.assertEqual(
+            response.data['error_msg'],
+            u'You can create up to {0} notes.'
+            u' You must remove some notes before you can add new ones.'.format(settings.MAX_ANNOTATIONS_PER_COURSE)
+        )
```
muzaffaryousaf: L391-402 are duplicated in `test_create_maximum_allowed`, can we refactor both tests to avoid the duplication. Even i think we don't need `test_create_maximum_allowed` test.

notesapi/v1/tests/test_views.py
```diff
@@ -346,6 +357,64 @@ def test_create_multirange(self):
...
+            self.assertEqual(
+                response.data['error_msg'],
+                u'You can create up to {0} notes.'
+                u' You must remove some notes before you can add new ones.'.format(settings.MAX_ANNOTATIONS_PER_COURSE)
+            )
+        else:
+            self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+    @patch('django.conf.settings.MAX_ANNOTATIONS_PER_COURSE', 5)
+    def test_create_maximum_other_user_course(self):
```
muzaffaryousaf: I would suggest `test_can_create_note_if_other_user_reached_limit`.

notesapi/v1/views.py
muzaffaryousaf: @muhammad-ammar  +1

notesapi/v1/views.py
```diff
@@ -275,20 +284,41 @@ def post(self, *args, **kwargs):  # pylint: disable=unused-argument
...
+                u'You can create up to {max_num_annotations_per_course} notes.'
+                u' You must remove some notes before you can add new ones.'
+            ).format(max_num_annotations_per_course=settings.MAX_ANNOTATIONS_PER_COURSE)
+            log.info(
+                u'Attempted to create more than %s annotations',
+                settings.MAX_ANNOTATIONS_PER_COURSE
+            )
+
+            # Gather metrics for New Relic so we can slice data in New Relic Insights
+            newrelic.agent.add_custom_parameter('notes.count', total_notes)
```
muzaffaryousaf: Why we added this here. I think it should be near create_note method because we have to monitor the number of notes that user is creating most of the time.

notesapi/v1/tests/test_views.py
```diff
@@ -346,6 +357,64 @@ def test_create_multirange(self):
...
+        self.assertFalse(hasattr(response, 'data'))
+
+        # if another user to tries to create note in first course it should succeed
+        token = get_id_token(TEST_OTHER_USER)
+        self.client.credentials(HTTP_X_ANNOTATOR_AUTH_TOKEN=token)
+        self.headers = {'user': TEST_OTHER_USER}
+        kwargs = {'user': TEST_OTHER_USER}
+        response = self._create_annotation(**kwargs)
+        self.assertFalse(hasattr(response, 'data'))
+
```
muzaffaryousaf: We can add test to check if the same user can create more notes in other course.

muzaffaryousaf:
@ehteshamkafeel I'm done with first pass, Let me know when address the feedback.

notesapi/v1/tests/test_views.py
ehteshamkafeel: it's there already

notesapi/v1/tests/test_views.py
ehteshamkafeel: I am removing the first test. `test_can_create_note_if_other_user_reached_limit ` is more than 40 characters which violates quality.

notesapi/v1/tests/test_views.py
muzaffaryousaf: +1

notesserver/settings/common.py
```diff
@@ -83,3 +83,6 @@
 )
 
 DEFAULT_NOTES_PAGE_SIZE = 25
+
+### Maximum number of allowed notes per course ###
+MAX_ANNOTATIONS_PER_COURSE = 500
```
symbolist: The other settings use the name notes, so lets call this `MAX_NOTES_PER_COURSE `. Also add "for each student" to the comment above.

notesapi/v1/views.py
```diff
@@ -275,20 +284,41 @@ def post(self, *args, **kwargs):  # pylint: disable=unused-argument
...
         """
-        if 'id' in self.request.data:
+        if not self.request.data or 'id' in self.request.data:
             return Response(status=status.HTTP_400_BAD_REQUEST)
 
         try:
+            total_notes = Note.objects.filter(
+                    user_id=self.request.data['user'], course_id=self.request.data['course_id']
+            ).count()
+            if total_notes >= settings.MAX_ANNOTATIONS_PER_COURSE:
```
symbolist: This should just be `>`.

notesapi/v1/tests/test_views.py
```diff
@@ -54,7 +55,7 @@ def setUp(self):
             "tags": ["pink", "lady"]
         }
 
-    def _create_annotation(self, **kwargs):
+    def _create_annotation(self, check_status=True, **kwargs):
```
symbolist: Instead of doing this, it would be cleaner to add parameters `expected_status=status.HTTP_201_CREATED, error_message=None`. And in the body do:

```
self.assertEqual(response.status_code, expected_status_code)
if error_message:
  self.assertEqual(response.data['error_msg'], error_message)

```

notesapi/v1/views.py
symbolist: Okay, no, this is correct.

symbolist:
@ehteshamkafeel Looks good. :+1:

muzaffaryousaf:
:+1:

PR #41, opened 2016-02-15
fix search when text and/or tags contain unicode data

muhammad-ammar:
## [TNL-4119](https://openedx.atlassian.net/browse/TNL-4119)

Search is not working if text or tags contain unicode characters. These changes will allow to search with unicode data.

When data is [indexed](https://github.com/edx/edx-notes-api/blob/master/notesapi/v1/search_indexes.py) from Note model into elasticsearch, a new [`data`](https://github.com/edx/edx-notes-api/blob/master/notesapi/v1/search_indexes.py#L15) attribute is added, this attribute is constructed by [concatenation](https://github.com/edx/edx-notes-api/blob/master/templates/search/indexes/v1/note_data.txt) of `note.text` and `note.tags` attributes. We search `text` or `tag` against this `data` attribute. So search was not working because `tags` list was not [serialized](https://github.com/edx/edx-notes-api/pull/41/files#diff-f89379700a3bdefbaf5244f886daa5aeR38) correctly during note creation.

[**Sandbox**](https://notes-pagination.sandbox.edx.org/courses/course-v1:edX+DemoX+Demo_Course/edxnotes/)

### Testing
- [x] Unit

### Reviewers
- [ ] Code review: @nedbat  
- [x] Code review: @robrap 
- [x] Code review: @symbolist 

### Post-review
- [ ] Squash commits

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ test: clean
 	./manage.py test --settings=$(test_settings) --with-coverage --with-ignore-docstrings \
 		--exclude-dir=notesserver/settings --cover-inclusive --cover-branches \
 		--cover-html --cover-html-dir=build/coverage/html/ \
-		--cover-xml --cover-xml-file=build/coverage/coverage.xml \
+		--cover-xml --cover-xml-file=build/coverage/coverage.xml --verbosity=2 \
 		$(foreach package,$(PACKAGES),--cover-package=$(package)) \
 		$(PACKAGES)
 
--- a/notesapi/v1/models.py
+++ b/notesapi/v1/models.py
@@ -35,6 +35,6 @@ class Note(models.Model):
 
         note_dict['ranges'] = json.dumps(ranges)
         note_dict['user_id'] = note_dict.pop('user', None)
-        note_dict['tags'] = json.dumps(note_dict.get('tags', list()))
+        note_dict['tags'] = json.dumps(note_dict.get('tags', list()), ensure_ascii=False)
 
         return cls(**note_dict)
--- a/notesapi/v1/tests/test_views.py
+++ b/notesapi/v1/tests/test_views.py
@@ -959,6 +959,26 @@ class AnnotationSearchViewTests(BaseAnnotationViewTests):
             start=start
         )
 
+    @ddt.unpack
+    @ddt.data(
+        {"text": u"Ammar محمد عمار Muhammad", "search": u"محمد عمار", "tags": [u"عمار", u"Muhammad", u"محمد"]},
+        {"text": u"Ammar محمد عمار Muhammad", "search": u"محمد", "tags": [u"محمد", u"Muhammad"]},
+        {"text": u"Ammar محمد عمار Muhammad", "search": u"عمار", "tags": [ u"ammar", u"عمار"]},
+        {"text": u"Muhammad Ammar", "search": u"عمار", "tags": [ u"ammar", u"عمار"]},
+        {"text": u"محمد عمار", "search": u"Muhammad", "tags": [ u"Muhammad", u"عمار"]}
+    )
+    @unittest.skipIf(settings.ES_DISABLED, "MySQL cannot do highlighting")
+    def test_search_unicode_text_and_tags(self, text, search, tags):
+        """
+        Verify that search works as expected with unicode and non-unicode text and tags.
+        """
+        self._create_annotation(text=text, tags=tags)
+
+        response = self._get_search_results(text=search)
+        self.assertEqual(response["total"], 1)
+        self.assertEqual(response["rows"][0]["text"], text)
+        self.assertEqual(response["rows"][0]["tags"], tags)
+
 
 @patch('django.conf.settings.DISABLE_TOKEN_CHECK', True)
 class AllowAllAnnotationViewTests(BaseAnnotationViewTests):


notesapi/v1/models.py
```diff
@@ -35,6 +35,6 @@ def create(cls, note_dict):
 
         note_dict['ranges'] = json.dumps(ranges)
         note_dict['user_id'] = note_dict.pop('user', None)
-        note_dict['tags'] = json.dumps(note_dict.get('tags', list()))
+        note_dict['tags'] = json.dumps(note_dict.get('tags', list()), ensure_ascii=False)
```
muhammad-ammar: `tags` in `note_dict` are already escaped with \uXXXX sequences. After applying `json.dumps` on tags list each unicode character is prefixed with another `\`, so `\u0441` becomes `\\u0441` which we don't want because if we write this data into elasticsearch, search will fail.

muhammad-ammar:
@nedbat @robrap @symbolist This is ready for review.

notesapi/v1/models.py
robrap: Does this mean there is bad data stored that needs to be fixed?

muhammad-ammar: @robrap Notes is not released yet so i don't expect any data to be fixed.

robrap: Got it.  Thanks.

muhammad-ammar:
@nedbat Please take a look.

notesapi/v1/tests/test_views.py
```diff
@@ -959,6 +959,26 @@ def test_pagination_page_start(self, page, start, previous_page, next_page):
...
         )
 
+    @ddt.unpack
+    @ddt.data(
+        {"text": u"Ammar محمد عمار Muhammad", "search": u"محمد عمار", "tags": [u"عمار", u"Muhammad", u"محمد"]},
+        {"text": u"Ammar محمد عمار Muhammad", "search": u"محمد", "tags": [u"محمد", u"Muhammad"]},
+        {"text": u"Ammar محمد عمار Muhammad", "search": u"عمار", "tags": [ u"ammar", u"عمار"]},
+        {"text": u"Muhammad Ammar", "search": u"عمار", "tags": [ u"ammar", u"عمار"]},
+        {"text": u"محمد عمار", "search": u"Muhammad", "tags": [ u"Muhammad", u"عمار"]}
+    )
```
robrap: I'm not clear on why all the different test cases and what they are testing, so it makes the test confusing to me.  For example, if we look at just the top 2 cases:
1. Is there anything special that happens with a search term broken by a space, or is this just a longer search term?
2. Does anything different happen when the tags are matched or not matched by the search string that we need to assert on?  Or, is the record just found if the search is matched in either the text or the tags?  If so, do you need more than 2 cases?  If you do, what is the english equivalent of describing each case?

Note: I think @jcdyer mentioned a clever way to use namedtuples (or something) to comment on ddt data.  If so, @jcdyer, can you send a link?

jcdyer: The namedtuple solution mostly solves a similar problem to this dict solution.  I didn't actually know ddt would unpack dicts.  One advantage of namedtuples is that it provides stricter type checking on the ddt data.  Another is that you can hack the default values to reduce verbosity in some cases.  The disadvantage is that you have to declare your namedtuple, which is a bit redundant.

To see how we did it, check out https://github.com/edx/edx-platform/blob/master/lms/djangoapps/course_blocks/transformers/tests/test_user_partitions.py#L285

robrap: I see.  Thanks @jcdyer, that makes sense.  
@muhammad-ammar I don't think it applies here.  I'd just like to see the cases simplified and clarified.  Maybe you could make the cases more clear by describing the strings in the strings?
```
{"text": u"Text with ascii and υηι¢σ∂є.", "search": u"υηι¢σ∂є", "tags": [u"unmatched"]},
```
I'm not sure, because I'm not clear on the cases you wish to test, which is my original point.

notesapi/v1/tests/test_views.py
```diff
@@ -959,6 +959,26 @@ def test_pagination_page_start(self, page, start, previous_page, next_page):
...
+    @unittest.skipIf(settings.ES_DISABLED, "MySQL cannot do highlighting")
```
robrap: Is this test running?  I could see any tests under Travis.  Can you post a link that helps me find it?  Thanks.

In addition to this, I wonder if we need a unicode related tests for Bokchoy.  Is it possible for something to be handled correctly in the model, but not make it to the page appropriately?

notesapi/v1/tests/test_views.py
muhammad-ammar: @robrap There is nothing special with these test cases, i am just verifying that if unicode search is working with different combinations of text and tags. As i mentioned earlier a new field is created during indexing by concatenation of text and tags fields and we do search against that concatenated field. So i am trying to check that search works with different text and tags values.

notesapi/v1/tests/test_views.py
muhammad-ammar: @robrap yes, it is running. I have increased the verbosity so now test names are show on console. You can check at https://travis-ci.org/edx/edx-notes-api/builds/109792097. Search `test_search_unicode_text_and_tags` in last two build on the above link.

I don't see any scenario in which wrong data will be passed to frontend. @nedbat Please share your thoughts.

robrap:
:+1: Thanks for your explanations @muhammad-ammar.

symbolist:
Lets merge this. :+1:

Makefile
```diff
@@ -13,7 +13,7 @@ test: clean
 	./manage.py test --settings=$(test_settings) --with-coverage --with-ignore-docstrings \
 		--exclude-dir=notesserver/settings --cover-inclusive --cover-branches \
 		--cover-html --cover-html-dir=build/coverage/html/ \
-		--cover-xml --cover-xml-file=build/coverage/coverage.xml \
+		--cover-xml --cover-xml-file=build/coverage/coverage.xml --verbosity=2 \
```
symbolist: Don't forget to remove this.

notesapi/v1/models.py
nedbat: I'm a little confused by this.  It sounds like we are using json.dumps to make a string, and then storing that string somewhere that doesn't interpret JSON?

symbolist: @nedbat Yes, the data is stored in MySQL and then passed onto ElasticSearch.

PR #44, opened 2016-03-09
Setting logging level to warning for ES.trace

muzaffaryousaf:
https://openedx.atlassian.net/browse/TNL-2441

### DIFF
--- a/notesserver/settings/logger.py
+++ b/notesserver/settings/logger.py
@@ -24,7 +24,6 @@ def get_logger_config(log_dir='/var/tmp',
     "edx_filename" is ignored unless dev_env is set to true since otherwise
     logging is handled by rsyslogd.
     """
-
     # Revert to INFO if an invalid string is passed in
     if local_loglevel not in ['DEBUG', 'INFO', 'WARNING', 'ERROR', 'CRITICAL']:
         local_loglevel = 'INFO'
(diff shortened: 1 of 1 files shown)


muzaffaryousaf:
Link to [es logging](https://github.com/elastic/elasticsearch-py/blob/fe6e9a7f5c832fd02847b99f3e2cb84292eb660b/docs/index.rst#logging).

muzaffaryousaf:
@feanil Would you mind to give it a look . Thanks.
@symbolist PTAL. Thanks.

symbolist:
This looks like a reasonable change. :+1:

notesserver/settings/logger.py
```diff
@@ -69,8 +68,8 @@ def get_logger_config(log_dir='/var/tmp',
             },
             "elasticsearch.trace": {
                 'handlers': handlers,
-                'level': 'DEBUG',
-                'propagate': False,
+                'level': 'WARNING',
+                'propagate': True,
```
feanil: Why do we need to change the propagation?  What other handlers are in the hierarchy of elasticsearch.trace?

muzaffaryousaf: Sorry, i think we don't need to propagate this to their parents.

muzaffaryousaf:
Closing in favor of this https://github.com/edx/edx-notes-api/pull/45

PR #52, opened 2017-02-01
Dockerize tests

fredsmith:
This moves travis tests into docker, running on 16.04. 

You can run: `make travis_up` to pull up the complete runtime environment, and `make travis_test` to run unit tests, and then `make travis_down` to tear down the running instances


.travis/run_tests.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash -xe
+. /edx/app/edx_notes_api/venvs/edx_notes_api/bin/activate
+
+cd /edx/app/edx_notes_api/edx_notes_api
+
+
+export ELASTICSEARCH_URL=http://es:9200/
```
feanil: We don't need this, we get them from the compose file.

fredsmith:
@muzaffaryousaf Hey, this test (test_search_course) is failing when I try to build this PR.  It seems to have ended up searching case-insensitive, when it was designed to do a case sensitive search.  I don't think I changed anything that should have affected this - I've tried with both elasticsearch 0.9.11 and 1.3.0

Makefile
```diff
@@ -1,16 +1,12 @@
...
 .PHONY: requirements
 
-validate: test.requirements test coverage
+include .travis/docker.mk
 
-ifeq ($(ESVER),-)
-test_settings = notesserver.settings.test_es_disabled
-else
-test_settings = notesserver.settings.test
-endif
```
muzaffaryousaf: @fredsmith I think different setting could be the reason for failure, not sure why we removed these settings.

muzaffaryousaf: NVM that's not the reason. ignore.

muzaffaryousaf:
@fredsmith Any idea how i can run these tests locally, same as before ?

fredsmith:
The test settings file now takes configuration via environment settings.  The defaults assume SQLite and elastic search on local host.  Then make validate, same as before.   You can run the Travis tests locally with `make travis_up; make travis_test; make travis_down`

.travis/docker-compose-travis.yml
```diff
@@ -0,0 +1,40 @@
...
+    container_name: db
+    command: mysqld --character-set-server=utf8 --collation-server=utf8_general_ci
+    environment:
+      MYSQL_ROOT_PASSWORD: ""
+      MYSQL_ALLOW_EMPTY_PASSWORD: "yes"
+      MYSQL_USER: "notes001"
+      MYSQL_PASSWORD: "secret"
+      MYSQL_DATABASE: "edx_notes_api"
+  es:
+    image: edxops/elasticsearch:0.9.13
```
muzaffaryousaf: @fredsmith  can you try with `1.4.2` too.

muzaffaryousaf: Ignore

muzaffaryousaf:
@fredsmith Earlier we were using sqlite(case sensitive) db for tests but for this PR we started to use mysql which is case insensitive by default that's why the test was failing. I think this test does't make sense since we don't explicitly make search case sensitive for course_id. I'd prefer to check case insensitivity for text parameter where we make it case insensitive explicitly.

muzaffaryousaf:
@fredsmith Build is green and i hope you're unblock on this now. Let me know if i can help. Thanks

fredsmith:
@muzaffaryousaf was that test searching elasticsearch or SQL?

PR #53, opened 2018-04-10
Notes API add endpoint to delete all notes for a user

yro:
### Description
Add 'delete all notes for user' endpoint for Notes API. Implements list view delete method.

### Testing
- [x] Unit tests

### Reviewers
- [x] Code review: @sanfordstudent 
- [ ] Code review: @doctoryes 

FYI: @edx/educator-neem

### Post-review
- [ ] Rebase and squash commits

### DIFF
--- a/notesapi/v1/tests/test_views.py
+++ b/notesapi/v1/tests/test_views.py
@@ -1,17 +1,16 @@
 # -*- coding: utf-8 -*-
-import urlparse
-import jwt
-import unittest
-import ddt
 from calendar import timegm
 from datetime import datetime, timedelta
-from mock import patch
+import ddt
+import jwt
+import unittest
+import urlparse
 
+from django.conf import settings
 from django.core.management import call_command
 from django.core.urlresolvers import reverse
-from django.conf import settings
 from django.test.utils import override_settings
-
+from mock import patch
 from rest_framework import status
 from rest_framework.test import APITestCase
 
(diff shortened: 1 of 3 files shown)


notesapi/v1/views.py
```diff
@@ -349,6 +348,19 @@ def post(self, *args, **kwargs):  # pylint: disable=unused-argument
         serializer = NoteSerializer(note)
         return Response(serializer.data, status=status.HTTP_201_CREATED, headers={'Location': location})
 
+    def delete(self, *args, **kwargs):  # pylint: disable=unused-argument
+        """
+        Delete all annotations for user_id
+
+        """
+        params = self.request.data
+        print(params)
```
sanfordstudent: Leftover print statement.

yro: 👍 Thank you.

sanfordstudent:
How is auth handled for Notes? If Notes can only be called from the LMS, this all seems good to me. If Notes doesn't already have that auth in place though, we should add something as part of this PR. The other GDPR retirement endpoints are all restricted to superusers only. If only the LMS can call into Notes, that's fine, because we can put the same restrictions on the endpoint that calls this endpoint (the cascading endpoint thing we discussed in grooming would actually make sense).

yro:
@sanfordstudent That's a really great question. There's no, or blisteringly out of date, documentation about this. I'll see if I can find something about it.

yro:
@sanfordstudent It looks like there's a token process (at least) from the LMS, I will do some more investigation, but it appears to be moderately secure.

yro:
@doctoryes Here's one of the PRs I was discussing with you yesterday. Thank you!

notesapi/v1/tests/test_views.py
```diff
@@ -508,6 +507,40 @@ def test_pagination_page_start(self, page, start, previous_page, next_page):
...
+        response = self._get_search_results()
+        self.assertEqual(response["total"], 0)
+
+        # Reattempt delete for User 1
+        response = self.client.delete(url, self.headers)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        # Delete for No User
+        del self.headers["user_id"]
+        response = self.client.delete(url, self.headers)
```
sanfordstudent: I see what's happening now. Django's `client.delete` takes args in the order `(path, data='', content_type='application/octet-stream', secure=False, **extra)` so `self.headers` is being passed as data to the request, not as headers.

yro: Makes perfect sense. Will update.

yro:
@sanfordstudent I believe I have updated this to reflect and assuage your concerns.

notesapi/v1/tests/test_views.py
```diff
@@ -508,6 +507,40 @@ def test_pagination_page_start(self, page, start, previous_page, next_page):
...
+
+        # Delete for No User
+        del self.payload["user_id"]
+        response = self.client.delete(url, headers=self.headers, data=self.payload)
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+        # Delete for User 2 with no notes
+        self.payload["user_id"] = TEST_OTHER_USER
+        response = self.client.delete(url, headers=self.headers, data=self.payload)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
doctoryes: Nit: There's a lot of code paths tested in this single test. I would have preferred to see each case split out to a separate test, even though separate setUp/tearDown might need to be performed.

yro: @doctoryes Sounds good - will do.

PR #58, opened 2018-06-06
Use an absolute path for the templates dir

pwnage101:
For some reason, this was necessary for the notes IDA in devstack,
otherwise it would just throw TemplateDoesNotExist.  It's so specific
that it shoudln't impact stage or prod which already seem to work
without this.

Also migrate TEMPLATE_DIRS variable to non-deprecated TEMPLATES
variable.

### DIFF
--- a/notesserver/settings/common.py
+++ b/notesserver/settings/common.py
@@ -80,9 +80,20 @@ CORS_ALLOW_HEADERS = (
     'x-annotator-auth-token',
 )
 
-TEMPLATE_DIRS = (
-    'templates',
-)
+# Base project path, where manage.py lives.
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
+
+TEMPLATES = [
+    {
+        'BACKEND': 'django.template.backends.django.DjangoTemplates',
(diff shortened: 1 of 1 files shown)


notesserver/settings/common.py
```diff
@@ -80,8 +80,9 @@
     'x-annotator-auth-token',
 )
 
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
 TEMPLATE_DIRS = (
-    'templates',
+    os.path.join(BASE_DIR, 'templates'),
```
doctoryes: Won't this addition affect other environments? Or is it expected that all templates in those environments would be found in the first `'templates'` entry?

pwnage101: It's still (edx) environment agnostic because the base URL is constructed by starting with the current location of this file, wherever it is.

If your question was about other installed apps, there's more magic happening in django's template-finding algorithm that I don't quite understand.  In particular, this did not break other installed apps, such as the API views from DRF (when I navigated to the API URLs from firefox), which would suggest that it still looks in app dirs.

I also went ahead and migrated to the new TEMPLATES setting since this setting has actually been deprecated since Django 1.8, and the new setting is slightly more explicit about things such as other apps.

PR #59, opened 2018-06-08
Update the expected user parameter from "user_id" to "user"

pwnage101:
This is necessary because the default permissions class for this repo
uses a hard-coded key "user" to check the user against the JWT token.

---

Related to https://github.com/edx/edx-platform/pull/18348

### DIFF
--- a/notesapi/v1/views.py
+++ b/notesapi/v1/views.py
@@ -203,17 +203,30 @@ class AnnotationListView(GenericAPIView):
                 Each page in the list contains 25 annotations by default. The page
                 size can be altered by passing parameter "page_size=<page_size>".
 
-                Http400 is returned if the format of the request is not correct.
+                HTTP 400 Bad Request: The format of the request is not correct.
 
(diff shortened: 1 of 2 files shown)


notesapi/v1/views.py
```diff
@@ -350,14 +350,13 @@ def post(self, *args, **kwargs):  # pylint: disable=unused-argument
 
     def delete(self, *args, **kwargs):  # pylint: disable=unused-argument
         """
-        Delete all annotations for user_id
-
+        Delete all annotations for user
```
doctoryes: Too bad it's not called `user_id` - I'm left wondering whether `user` is a user ID, username, object, etc....

pwnage101: I'd argue user_id was potentially worse because it's also not *the* user id since it's actually a hash of the user id.  I really should just clarify the docstring here.

doctoryes: Agreed - that's worse!

PR #60, opened 2018-08-02
Upgrade Django and all its dependencies.

attiyaIshaque:
### Description

Update Django. 
Older version: 1.8.7
New version: 1.11.5

This also includes a small change to cover implementation of url patterns. 

### Reviewers
- [ ] @ssemenova 
- [ ] @fysheets 
- [ ] @awaisdar001  
 
### Post-review
- [ ] Rebase and squash commits

### DIFF
--- a/notesapi/urls.py
+++ b/notesapi/urls.py
@@ -1,6 +1,5 @@
-from django.conf.urls import patterns, url, include
+from django.conf.urls import url, include
 
-urlpatterns = patterns(
-    '',
+urlpatterns = [
     url(r'^v1/', include('notesapi.v1.urls', namespace='v1')),
-)
+]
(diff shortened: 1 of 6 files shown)


awaisdar001:
Looks good to me!

Question: Do we need a release tag for this as well?

attiyaIshaque:
@awaisdar001 we don't need a release tag for this. (i think so) :)

nedbat:
This has been cherry-picked onto Hawthorn.

attiyaIshaque:
@nedbat do you have any idea when will this fix goes live?

nedbat:
@attiyaIshaque live in edX production? I have no idea.  We are going to ship Hawthorn this week, and people will start running it.

PR #61, opened 2018-08-10
Add metrics for requests.

robrap:
ARCH-193

### DIFF
--- a/notesserver/settings/common.py
+++ b/notesserver/settings/common.py
@@ -38,9 +38,11 @@ RESULTS_MAX_SIZE = 250
 ROOT_URLCONF = 'notesserver.urls'
 
 MIDDLEWARE_CLASSES = (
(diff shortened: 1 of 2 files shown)


notesserver/settings/common.py
```diff
@@ -38,9 +38,11 @@
 ROOT_URLCONF = 'notesserver.urls'
 
 MIDDLEWARE_CLASSES = (
+    'edx_django_utils.cache.middleware.RequestCacheMiddleware',
     'corsheaders.middleware.CorsMiddleware',
     'django.middleware.common.CommonMiddleware',
     'django.middleware.csrf.CsrfViewMiddleware',
+    'edx_rest_framework_extensions.middleware.RequestMetricsMiddleware',
```
robrap: Reviewer FYI: Since there is no Session Middleware this was failing with the TieredCacheMiddleware, so I dropped it since that is only used for testing.

PR #67, opened 2018-10-17
Rename DELETE /api/v1/annotations/ to POST /api/v1/retire_annotations

pwnage101:
This is a mitigation to the issue raised in LEARNER-6532 where
apparently buggy annotator.js code is calling the DELETE endpoint
without providing an annotation id.  Our annotator-compatible API
deviates from the documented annotator API in such a way that NOT
providing the annotation id implies delete (retire) all annotations.
    
The correct fix should be to additionally protect this API endpoint by
only authorizing the LMS, but renaming the endpoint is a sufficient
mitigation since annotator.js is far less likely to accidentally call
it.

### DIFF
--- a/notesapi/v1/tests/test_views.py
+++ b/notesapi/v1/tests/test_views.py
@@ -517,19 +517,19 @@ class AnnotationListViewTests(BaseAnnotationViewTests):
         response = self._get_search_results()
         self.assertEqual(response["total"], 3)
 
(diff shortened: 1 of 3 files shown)


pwnage101:
@waheedahmed yes we do, but that PR will come later (today, hopefully).  The final version of this code in edx-notes-api will look different still, with stricter auth, so that should come first.

PR #87, opened 2019-01-18
Enable PII annotation checking

doctoryes:
https://openedx.atlassian.net/browse/PLAT-2386

Annotate the single Django model. Add commands for CI to run checking in Travis.

### DIFF
--- a/.annotation_safe_list.yml
+++ b/.annotation_safe_list.yml
@@ -1,0 +1,10 @@
+auth.Group:
+  ".. no_pii::" : "No PII"
+auth.Permission:
+  ".. no_pii::" : "No PII"
+auth.User:
+  ".. pii::": "username and email address"
+  ".. pii_types::" : name, other
+  ".. pii_retirement::" : local_api
+contenttypes.ContentType:
+  ".. no_pii::": "No PII"
--- a/.pii_annotations.yml
+++ b/.pii_annotations.yml
@@ -1,0 +1,15 @@
+source_path: ./
+report_path: pii_reports
+safelist_path: .annotation_safe_list.yml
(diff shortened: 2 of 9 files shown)


.annotation_safe_list.yml
```diff
@@ -0,0 +1,10 @@
+auth.Group:
+  ".. no_pii::" : "No PII"
+auth.Permission:
+  ".. no_pii::" : "No PII"
+auth.User:
+  ".. pii::": "username and email address"
```
pwnage101: For some reason, I didn't realize that if we don't do the tag prefix/suffix thing that bmez suggested a while back, then the safelist keys would need to use the fully marked up sphinx-type variant of the tag :(

.annotations
```diff
@@ -0,0 +1,15 @@
+source_path: ../
```
pwnage101: Should this be `../` or `./`?  wouldn't `../` search other files outside the repo?

doctoryes: Good catch! It doesn't affect this CI, since the Docker container does this checking from this directory:
`/edx/app/edx_notes_api/edx_notes_api`
But the checking *should* be done from `./` - will fix.

pwnage101:
Given that other teams are interested in using code-annotations for other things besides PII, I wonder if we should get in the habit of not calling the config file `.annotations`.  For notes it's no big deal, but it'll set a precedence.

doctoryes:
> Given that other teams are interested in using code-annotations for other things besides PII, I wonder if we should get in the habit of not calling the config file `.annotations`. For notes it's no big deal, but it'll set a precedence.

Great point - I'll rename it to `.pii_annotations`

PR #88, opened 2019-01-22
make pii_report directory name consistent

pwnage101:
The pii_* make targets use "pii_report" instead of "pii_reports".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 .coverage
 .tox
 coverage/
+pii_report/
 
 # Sqlite Database
 *.db
--- a/.pii_annotations.yml
+++ b/.pii_annotations.yml
@@ -1,5 +1,5 @@
 source_path: ./
-report_path: pii_reports
+report_path: pii_report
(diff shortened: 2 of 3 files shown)


doctoryes:
Did you want to add `pii_report` to `.gitignore` as well?

doctoryes:
Also, the `Makefile` has the name defined separately - for the purpose of re-creating the correct directory.

pwnage101:
> Did you want to add pii_report to .gitignore as well?

I'll add that too.  This PR was initially generated with github's web UI so it wasn't possible the first time around.

> Also, the Makefile has the name defined separately - for the purpose of re-creating the correct directory.

Can you elaborate?  Was the directory name in the makefile supposed to be different than the one in the config file?

doctoryes:
No - the directory name was supposed to be the same. I passed the name on the cmd-line in `Makefile` so it could be easily seen that the cleaning of the PII reports was done in the same directory as the generated reports.

pwnage101:
Forgot I had this PR lingering around.

PR #91, opened 2019-02-22
Add username replacement API

tuchfarber:
New API replaces all copies of username in this service with a new
username. Requires user be in a username_replacement_admin group. Should
only be ran as a larger job to update usernames across all services,
otherwise the system will be left in a broken state for those users.


tuchfarber:
We don't actually store anything in auth_user and the notes only have a user_id, so this is safe to close.

PR #85, opened 2018-12-27
[BB-724] Disable default BasicAuthentication 

xitij2000:
By default, BasicAuthentication is included as one of the ``DEFAULT_AUTHENTICATION_CLASSES``, and this is causing a problem with our setup which has the notes-api running behind HAProxy. 

When the platform is set up to require basic authentication and the header for the same is added to all API requests automatically, edx-notes-api tries to parse the header and match it an existing user account. This causes API requests to fail with an invalid username/password error. 

This PR explicitly sets the ``DEFAULT_AUTHENTICATION_CLASSES`` to exclude BasicAuthentication, since that is not used by this API. 

**Testing instructions**:

1. Install edx-notes-api
2. Set it up behind Nginx or another reverse proxy that requires authentication.
3. Try to access APIs that should work without an Authentication header. 
4. You should get an "invalid username/password error" with the master branch, but none with this branch. 

**Reviewers**
- [ ] @giovannicimolin 
- [ ] edX reviewer[s] TBD

PR #93, opened 2019-05-15
Revert "[BB-724] Disable default BasicAuthentication "

bderusha:
Reverts edx/edx-notes-api#85

### DIFF
--- a/notesserver/settings/common.py
+++ b/notesserver/settings/common.py
@@ -66,9 +66,6 @@ WSGI_APPLICATION = 'notesserver.wsgi.application'
 LOGGING = get_logger_config()
 
 REST_FRAMEWORK = {
-    'DEFAULT_AUTHENTICATION_CLASSES': [
-        'rest_framework.authentication.SessionAuthentication',
-    ],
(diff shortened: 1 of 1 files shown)


ormsbee:
@xitij2000: Heads up on this revert -- when we finally released a new version of Notes with this change in it, we got a slew of auth related errors.

ormsbee:
@bderusha: Can you paste in the errors you were seeing?

xitij2000:
@ormsbee Perhaps we can make this configurable, and keep ``BasicAuthentication`` enabled by default? Adding this option will break the API in our use case. 

I'm not sure I understand why removing it would cause issues though, since the API doesn't authenticate using basic auth.

ormsbee:
@xitij2000: Sorry for the slow replies. I've been sick as of late.

It looks like this was a red herring on our part, so I'm reverting the revert (i.e. keeping your contribution from #85). Apologies for the trouble.

xitij2000:
@ormsbee No worries. I hope you're feeling better now!

PR #94, opened 2019-05-22
Revert "Revert "[BB-724] Disable default BasicAuthentication ""

ormsbee:
Reverts edx/edx-notes-api#93 (which reverted #85). The changes in #85 were originally reverted because we thought they were responsible for an auth related error prevented us from deploying the notes API. However, we continued to see the same issue after reverting. Since this wasn't the root of the problem, I'm reverting the revert (i.e. putting the original contribution back into master).


ormsbee:
FYI @xitij2000

PR #98, opened 2019-08-08
Update openedx.yaml

feanil:
Update owner to Bill

### DIFF
--- a/openedx.yaml
+++ b/openedx.yaml
@@ -7,4 +7,4 @@ oeps:
     state: true
   oep-7: true  # Python 3
 openedx-release: {ref: master}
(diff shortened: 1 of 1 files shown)


feanil:
This needs an owner so that it can keep getting regular security updates.  Bill was interested in owning this.  Starting  a conversation.

bderusha:
I agree to own this service as I consider it under "active development" from the Platform Health squad while we transition it to run on k8s infrastructure

feanil:
@bderusha do you want to wait for a response from @e0d or should we just merge this?

bderusha:
I am comfortable merging this.  If @e0d thinks this is the wrong choice we can always update

PR #106, opened 2019-08-22
Dockerfile and helmchart

coryleeio:


coryleeio:
helm dependency build
helm package notes
helm install notes-0.1.0.tgz --name notes -f notes/values.yaml

coryleeio:
WIP
https://github.com/edx/edx-notes-api/pull/106

adzuci:
Cool tricks:
```
docker build . -t edxops/notes:latest && k delete pod $(k get pods | grep notes | head -n1 | awk '{print $1}')
```

notesserver/settings/common.py
```diff
@@ -2,7 +2,7 @@
 import os
 import sys
 
-from notesserver.settings.logger import get_logger_config
+from notesserver.settings.logger import build_logging_config
```
bderusha: Unused import?

notesserver/settings/yaml_config.py
```diff
@@ -43,3 +42,6 @@
 if ES_DISABLED:
     HAYSTACK_CONNECTIONS = {}
     INSTALLED_APPS.remove('haystack')
+
+LOGGING = build_logging_config()
```
bderusha: Ah, used here.  Can we move the import here instead?

Dockerfile
```diff
@@ -0,0 +1,39 @@
+FROM ubuntu:bionic
+MAINTAINER devops@edx.org
+
+RUN apt-get update && apt-get upgrade -qy && apt-get install language-pack-en locales git python2.7 python-pip libmysqlclient-dev -qy && \
```
bderusha: Would be great to list each dependency and what it does / why we need it.

```
git; Used to pull in particular requirements from github rather than pypi
```

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+# this prevents the image cache from busting unless the dependencies have changed.
+COPY requirements/base.txt /edx/app/notes/requirements/base.txt
+
+# Dependencies are installed as root so they cannot be modified by the application user.
+RUN pip install -r requirements/base.txt
+
+RUN mkdir -p /edx/var/log
+
+# Code is owned by root so it cannot be modified by the application user.
+# So we copy it before changing users.
```
bderusha: Where is this copy cmd?

helmcharts/notes/templates/_helpers.tpl
```diff
@@ -0,0 +1,67 @@
+{{/* vim: set filetype=mustache: */}}
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "notes.name" -}}
+{{- default .Chart.Name .Values.app.nameOverride | trunc 63 | trimSuffix "-" -}}
```
bderusha: Can we indent these blocks (and the rest of this file)?

helmcharts/notes/templates/deployment.yaml
```diff
@@ -0,0 +1,86 @@
...
+          # Expose application secret contents as a volume in /edx/etc
+          volumeMounts:
+            - mountPath: "/edx/etc"
+              name: {{ include "notes.secretName" . }}
+              readOnly: true
+          resources:
+            {{- toYaml .Values.app.resources | nindent 12 }}
+      initContainers:
+      - name: init-elasticsearch
+        image: busybox:1.29.3
```
bderusha: Just did some reading on the BusyBox image.  Would totally be worth a comment somewhere about what this image is and why it is useful.

helmcharts/notes/templates/secrets.yaml
```diff
@@ -0,0 +1,66 @@
...
+    app: {{ template "mysql.fullname" $ }}
+    chart: "{{ $.Chart.Name }}-{{ $.Chart.Version }}"
+    release: "{{ $.Release.Name }}"
+    heritage: "{{ $.Release.Service }}"
+type: Opaque
+data:
+  ca.pem: {{ .ca | b64enc }}
+  server-cert.pem: {{ .cert | b64enc }}
+  server-key.pem: {{ .key | b64enc }}
+{{- end }}
```
bderusha: Indenting might be nice here, but may also look bad with the rest of the YAML format.   If so, a comment referencing which block the `end` is ending would be helpful. 

I think the code blocks map to the following:
```
{{- end }} # end range .Values.mysql.ssl.certificates
{{- end }} # end if .Values.mysql.ssl.certificates
{{- end }} # end if .Values.mysql.ssl.enabled
{{- end }} # end if .Values.mysql.enabled
```

helmcharts/notes/templates/tests/test-connection.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  name: "{{ include "notes.fullname" . }}-test-connection"
+  labels:
+{{ include "notes.labels" . | indent 4 }}
+  annotations:
+    "helm.sh/hook": test-success
+spec:
+  containers:
+    - name: wget
+      image: busybox
```
bderusha: In other specs we specify a version of the busybox image.  Should we specify here as well?

helmcharts/notes/values.yaml
```diff
@@ -0,0 +1,127 @@
...
+    USERNAME_REPLACEMENT_WORKER: username_replacement_service_worker
+    LOG_SETTINGS_LOG_DIR: '/var/tmp'
+    LOG_SETTINGS_LOGGING_ENV: 'no_env'
+    LOG_SETTINGS_DEV_ENV: True
+    LOG_SETTINGS_DEBUG: True
+    LOG_SETTINGS_LOCAL_LOGLEVEL: 'INFO'
+    LOG_SETTINGS_EDX_FILENAME: "edx.log"
+    LOG_SETTINGS_SERVICE_VARIANT: 'edx-notes-api'
+
+# See the Elasticsearch chart provider's website in requirements.yaml for a full list of possible overrides
```
bderusha: Link?

helmcharts/notes/values.yaml
```diff
@@ -0,0 +1,127 @@
...
+elasticsearch:
+  enabled: true
+
+  image:
+    repository: "elasticsearch"
+    tag: "1.5.2"
+    pullPolicy: "IfNotPresent"
+
+# See the Mysql chart provider's website in requirements.yaml for a full list of possible overrides
```
bderusha: Link?

helmcharts/notes/values.yaml
```diff
@@ -0,0 +1,127 @@
...
+mysql:
+  enabled: true
+  
+  image: mysql
+  imageTag: 5.6.45
+  existingSecret: notes-mysql-secrets
```
bderusha: I have so much to learn :guitar:

notesserver/docker_gunicorn_configuration.py
```diff
@@ -0,0 +1,45 @@
+"""
+gunicorn configuration file: http://docs.gunicorn.org/en/develop/configure.html
```
bderusha: Let's ship this, verify it, and move to gunicorn 20 ASAP :D

notesserver/settings/dev.py
```diff
@@ -13,4 +13,14 @@
     }
 }
 
-LOGGING = get_logger_config(debug=DEBUG, dev_env=True, local_loglevel='DEBUG')
+LOG_SETTINGS = {
```
bderusha: I don't see how or where this is used. Can we kill it?  Why or why not?

notesserver/settings/devstack.py
```diff
@@ -1,12 +1,22 @@
...
 
 from .common import *
 
 DEBUG = True
 
 ALLOWED_HOSTS = ['*']
 
-LOGGING = get_logger_config(debug=DEBUG, dev_env=True, local_loglevel='DEBUG')
+
+LOG_SETTINGS = {
```
bderusha: I don't see how or where this is used. Can we kill it?  Why or why not?

notesserver/settings/yaml_config.py
```diff
@@ -22,10 +22,9 @@
 CONFIG_ROOT = path(EDXNOTES_CONFIG_ROOT)
 
 with open(CONFIG_ROOT / "edx_notes_api.yml") as yaml_file:
-    config_from_yaml = yaml.load(yaml_file)
+    config_from_yaml = yaml.safe_load(yaml_file)
```
bderusha: :guitar:

Dockerfile
coryleeio: Added

Dockerfile
coryleeio: Moved it - ty

coryleeio: I verified file ownership is still root, the user directive is not taking place until run time, files are still owned by root despite the reorder.

helmcharts/notes/templates/_helpers.tpl
coryleeio: This file with the exception of the last two definitions:
notes.requiredSecretName
notes-mysql.secretName

was just generated by helm create.

I believe indenting would add tabs to the places it is used and is not typically done in the stable charts which I tried to mirror as much as possible.  Here is an example:
https://github.com/helm/charts/blob/master/stable/ambassador/templates/_helpers.tpl

Does that change your opinion here?

coryleeio: Indenting is typically added to the template via piping to '| indent 6' for example

helmcharts/notes/templates/deployment.yaml
coryleeio: Added

helmcharts/notes/templates/secrets.yaml
coryleeio: Same comment about indenting from above. The secret definition was more or less taken from 
https://github.com/helm/charts/blob/master/stable/mysql/templates/secrets.yaml

I agree with you that adding the comments would make it more readable though.  Updated.

helmcharts/notes/templates/tests/test-connection.yaml
coryleeio: done

helmcharts/notes/values.yaml
coryleeio: added

helmcharts/notes/values.yaml
coryleeio: added

helmcharts/notes/values.yaml
coryleeio: I added a comment explaining this because it was actually kind of hard to sort through and I feel like it's unclear now that you called it out the comment reads:

`  # We create the secret ourselves in templates/secrets.yaml
  # Defining the name of the secret in templates/_helpers.tpl
  # and use the existing secret directive to tell the mysql chart not to create
  # secrets to use ours instead.
  # this is done so we can use the secret from our values.yaml since mysql and 
  # the django application need to have the same DB password.
  # This is a common practice in the helm stable charts 
  # https://github.com/helm/charts/tree/master/stable
  # as these charts are often used to compose higher level charts.
  # but should be able to run on their own out of the box.`

notesserver/docker_gunicorn_configuration.py
coryleeio: agreed

notesserver/settings/common.py
coryleeio: Good catch - thanks

notesserver/settings/devstack.py
coryleeio: The devstack one is used here:
https://github.com/edx/devstack/blob/master/docker-compose.yml#L177-L196

It is my belief that it should be using the yaml config and should load a yaml from that repo instead to reduce the number of codepaths, but that is currently not the case.

notesserver/settings/dev.py
coryleeio: Ah good catch.  This was part of my iterating and I left it here by accident.

coryleeio: removed

notesserver/settings/devstack.py
coryleeio: Good catch - this was a precursor to the broken out variables I am now using - removed.

notesserver/settings/yaml_config.py
coryleeio: :+1:

notesserver/settings/yaml_config.py
coryleeio: Yes that seems smart. Updated.

Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+
+
+# Packages installed:
+# git; Used to pull in particular requirements from github rather than pypi, 
+# and to check the sha of the code checkout.
+
+# language-pack-en locales; ubuntu locale support so that system utilities have a consistent
+# language and time zone.
+
+# python; ubuntu doesnt ship why python, so this is the python we will use to run the application
```
bderusha: why => with

Dockerfile
bderusha: Can we update/move the comment so it isn't as confusing

helmcharts/notes/templates/_helpers.tpl
bderusha: If this is standard, that is fine.  I don't have a strong preference, especially where it is easy to reason about the blocks.

PR #118, opened 2019-11-13
update notes docker target

fredsmith:


.travis/docker.mk
```diff
@@ -21,8 +21,10 @@ travis_down: ## Stop and remove containers and other resources created by `travi
...
 	echo "$$DOCKER_PASSWORD" | docker login -u "$$DOCKER_USERNAME" --password-stdin
 
-docker_push: docker_auth ## push to docker hub
-	docker build . -f Dockerfile -t edxops/edx-notes-api
-	docker push 'edxops/edx-notes-api'
-	docker build . -f Dockerfile.newrelic -t 'edxops/edx-notes-api:newrelic'
-	docker push 'edxops/edx-notes-api:newrelic'
+docker_build:
+	docker build . -f Dockerfile -t 'edxops/notes:latest'
+	docker build . -f Dockerfile.newrelic -t 'edxops/notes:newrelic-latest'
```
adzuci: Is the idea that we would use this tag and would push a non-newrelic image to latest for non edX use cases?

PR #124, opened 2019-12-03
Use a multistage build to avoid the extra dockerfile

coryleeio:


coryleeio:
You can build app only like so:
docker build --target=app .

build app with NR like so:
docker build .

adzuci:
> You can build app only like so:
> docker build --target=app .
> 
> build app with NR like so:
> docker build .

It may be worth putting this in a doc somewhere.

PR #138, opened 2020-02-25
BOM-1331

awais786:
replace nose with pytest.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,12 +6,7 @@ include .travis/docker.mk
 validate: test.requirements test
 
 test: clean
-	./manage.py test --settings=notesserver.settings.test --with-coverage --with-ignore-docstrings \
-		--exclude-dir=notesserver/settings --cover-inclusive --cover-branches \
-		--cover-html --cover-html-dir=build/coverage/html/ \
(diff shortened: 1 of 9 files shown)


pytest.ini
```diff
@@ -0,0 +1,4 @@
+[pytest]
+DJANGO_SETTINGS_MODULE = notesserver.settings.test
+addopts = --cov=notesapi --cov-report term --cov-config=.coveragerc --no-cov-on-fail
```
Ayub-Khan: Need to generate cov for both "notesserver notesapi"

Makefile
```diff
@@ -6,12 +6,7 @@ include .travis/docker.mk
 validate: test.requirements test
 
 test: clean
-	./manage.py test --settings=notesserver.settings.test --with-coverage --with-ignore-docstrings \
-		--exclude-dir=notesserver/settings --cover-inclusive --cover-branches \
```
jmbowman: We should probably update `.coveragrc` to still calculate branch coverage; see the `branch` option under https://coverage.readthedocs.io/en/coverage-5.0.3/config.html#run

awais786: done

PR #140, opened 2020-03-02
BOM-1344 :  Django 2.2 updates in edx-notes-api

mzulqarnain1:
Related JIRA : [https://openedx.atlassian.net/browse/BOM-1344](https://openedx.atlassian.net/browse/BOM-1344)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,10 +20,14 @@ after_success:
 - codecov
 env:
   matrix:
-  - DOCKER_USERNAME=edxbuilder
+  - TOXENV=django111
+  - TOXENV=django20
+  - TOXENV=django21
+  - TOXENV=django22
   global:
+    - DOCKER_USERNAME=edxbuilder
     # encrypted DOCKER_PASSWORD
-    secure: OgsQf8hp40s+E2btyl+RHcdgjqx9gZ4phlvm2tQ8Oy9mCO6hAy2XCjL1m2bidaFzwSjZXDZ7BhjD3yXpW2B+zP/yJNVKmdPA5E+D7YgFPdlOv7Wb+t0Z083Wbwqk7ZZgxLCcMLWf5vTs/QkPVAXbU8wJvQQC2l0EKESFnYkzNAI=
+    - secure: OgsQf8hp40s+E2btyl+RHcdgjqx9gZ4phlvm2tQ8Oy9mCO6hAy2XCjL1m2bidaFzwSjZXDZ7BhjD3yXpW2B+zP/yJNVKmdPA5E+D7YgFPdlOv7Wb+t0Z083Wbwqk7ZZgxLCcMLWf5vTs/QkPVAXbU8wJvQQC2l0EKESFnYkzNAI=
 deploy:
   provider: script
   script: make docker_push
--- a/.travis/docker.mk
+++ b/.travis/docker.mk
@@ -7,10 +7,10 @@ travis_start: ## Start containers stopped by `travis_stop`
 	docker-compose -f .travis/docker-compose-travis.yml start
 
 travis_test: ## Run tests on Docker containers, as on Travis CI
-	docker exec -it edx_notes_api env TERM=$(TERM) /edx/app/edx_notes_api/edx_notes_api/.travis/run_tests.sh
+	docker exec -e TERM=$(TERM) -e TOXENV=$(TOXENV) -it edx_notes_api /edx/app/edx_notes_api/edx_notes_api/.travis/run_tests.sh
 
 travis_pii_check: ## Run pii annotations checker on Docker containers, as on Travis CI
-	docker exec -it edx_notes_api env TERM=$(TERM) /edx/app/edx_notes_api/edx_notes_api/.travis/run_pii_checker.sh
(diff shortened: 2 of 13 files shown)


requirements/base.txt
```diff
@@ -12,7 +12,6 @@ django-cors-headers==3.2.1  # via -r requirements/base.in (line 7)
 django-haystack==2.8.0    # via -r requirements/base.in (line 5)
 django-rest-swagger==2.1.2  # via -r requirements/base.in (line 4)
 django-waffle==0.20.0     # via edx-django-utils, edx-drf-extensions
-django==1.11.28           # via -r requirements/base.in (line 1), django-cors-headers, django-haystack, djangorestframework, edx-django-release-util, edx-django-utils, edx-drf-extensions, rest-condition
```
jmbowman: Why was the django dependency removed from this file?  Should have only been removed from `test.txt`.

tox.ini
```diff
@@ -0,0 +1,16 @@
+[tox]
+envlist = {py36}-django{111,20,21,22}
```
jmbowman: Shouldn't this be `py35` to match the current production environment?

.travis/docker.mk
```diff
@@ -7,7 +7,7 @@ travis_start: ## Start containers stopped by `travis_stop`
 	docker-compose -f .travis/docker-compose-travis.yml start
 
 travis_test: ## Run tests on Docker containers, as on Travis CI
-	docker exec -it edx_notes_api env TERM=$(TERM) /edx/app/edx_notes_api/edx_notes_api/.travis/run_tests.sh
+	$(TOX)pytest
```
jmbowman: Can you set `TOXENV` (like `TERM` above) and update `run_tests.sh` to respect it instead?  That would still let you run the tests in the Docker container, and presumably avoid the need to switch to the `test_es_disabled` settings.

.travis.yml
```diff
@@ -19,13 +19,17 @@ after_success:
...
 env:
-  matrix:
+- TOXENV=django111
+- TOXENV=django20
+- TOXENV=django21
+- TOXENV=django22
+matrix:
   - DOCKER_USERNAME=edxbuilder
-  global:
+global:
```
jmbowman: These `matrix` and `global` keys need to stay inside `env`.  The environment variables in `global` are set for all workers, the ones in `matrix` define different workers (see https://docs.travis-ci.com/user/environment-variables/#global-variables for details).  So I think what you want here is to move `DOCKER_USERNAME` into `global` and put the different `TOXENV` values under `matrix`.

tox.ini
mzulqarnain1: changed.

.travis/docker.mk
```diff
@@ -7,10 +7,10 @@ travis_start: ## Start containers stopped by `travis_stop`
 	docker-compose -f .travis/docker-compose-travis.yml start
 
 travis_test: ## Run tests on Docker containers, as on Travis CI
-	docker exec -it edx_notes_api env TERM=$(TERM) /edx/app/edx_notes_api/edx_notes_api/.travis/run_tests.sh
+	docker exec -it edx_notes_api env TERM=$(TERM) | TOX_ENV_DIR=$(TOX_ENV_DIR) /edx/app/edx_notes_api/edx_notes_api/.travis/run_tests.sh
```
mzulqarnain1: Hey @jmbowman Tox documentation says that it injects [`TOX_ENV_DIR` into the environment](https://tox.readthedocs.io/en/latest/config.html) and is working fine when I run tests locally with Tox, but on Travis it always has an empty value. Can you help me in figuring out this problem?

jmbowman: You're not inside tox yet at this point; you're in a call to `make` directly from the Travis worker VM.  I think you want to use `TOXENV=$(TOXENV)` instead, since that's what was set as an environment variable by `.travis.yml`.

mzulqarnain1: I have added **`TOXENV`** variable instead now, and the environment being created is correct, but the build is failing when pytest command is executed with following trace.
```
tox -- python -Wd -m pytest
make: docker: Command not found
.travis/docker.mk:10: recipe for target 'travis_test' failed
make: *** [travis_test] Error 127
```
Can you help with this problem?

jmbowman: Currently `make travis_test` calls tox, which then calls `make travis_test`.  I think instead you want tox to call `make validate`, which is what `make travis_test` originally did after all the docker and virtualenv setup.

mzulqarnain1: aah, Thanks.

mzulqarnain1:
Hey @jmbowman , This PR is good to go now. One thing to note is that we still need to test it with `test_es_disabled` settings file, otherwise tests start failing.

jmbowman:
I suspect the cause of the failing tests is that the environment variables from the docker-compose file aren't being passed into the tox environment; you'll probably need to add these to the `passenv` section: https://github.com/edx/edx-notes-api/blob/master/.travis/docker-compose-travis.yml#L32-L40 .  Without those it's looking for ElasticSearch on localhost instead of the separate container, etc.

mzulqarnain1:
> I suspect the cause of the failing tests is that the environment variables from the docker-compose file aren't being passed into the tox environment; you'll probably need to add these to the `passenv` section: https://github.com/edx/edx-notes-api/blob/master/.travis/docker-compose-travis.yml#L32-L40 . Without those it's looking for ElasticSearch on localhost instead of the separate container, etc.

@jmbowman You were right. It's good to go now.

PR #151, opened 2020-04-10
travis git PR fixes

fredsmith:


.travis.yml
```diff
@@ -25,15 +25,15 @@ env:
   - TOXENV=django21
   - TOXENV=django22
   global:
```
adzuci, line 27: What are the two secure lines in this block (Docker and GitHub?). Looks like the Travis CLI changed the format so it might be helpful to add comments back in.

PR #154, opened 2020-04-27
remove tox

fredsmith:


.travis.yml
```diff
@@ -19,11 +19,6 @@ after_success:
 - docker exec notes /edx/app/notes/notes/.travis/run_coverage.sh
 - codecov
 env:
-  matrix:
-  - TOXENV=django111
-  - TOXENV=django20
-  - TOXENV=django21
-  - TOXENV=django22
```
adzuci, line 26: Should we also remove tox from test.requirements?

jmbowman:
@fredsmith The Arbisoft team is wondering what the reason for removing this was; is it just to reduce the number of Jenkins workers?  They were about to remove the entries for Django 1.11 through 2.1, leaving it just running 2.2 (with 3.0 to be added soon to start preparing for 3.2 compatibility).

awais786:
@fredsmith any update on this ?

jmbowman:
@awais786 I just checked with Fred, he said that some of the configurations were failing and tox was removed to get the Docker image build working again.  Should be fine to add it back as long as tests are passing and we only test the configurations we're actively using or about to switch to.

PR #156, opened 2020-04-27
Setting auth table to auth.User instead of core.User since core.User doesn't exist

coryleeio:


notesserver/settings/common.py
```diff
@@ -7,6 +7,7 @@
 DISABLE_TOKEN_CHECK = False
 USE_TZ = True
 TIME_ZONE = 'UTC'
+AUTH_USER_MODEL = 'auth.User'
```
adzuci, line 10: I'm not sure if I like that this would make notes more different than ever other IDA?

coryleeio: we dont have core_user tables in prod, we have auth_user tables though.

adzuci: Hmm, @jmbowman mentioned that this should default to auth.User, so this may not fix things if that's true: https://docs.djangoproject.com/en/2.2/ref/settings/#auth-user-model

coryleeio: It fixed it locally and I also verified that the old value was auth.User 2 months ago

PR #165, opened 2020-05-13
Adding py3.8 support

awais786:
https://openedx.atlassian.net/browse/BOM-1628
Adding py3.8 support.

Tox removed from travis in this https://github.com/edx/edx-notes-api/pull/154
Added back tox.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,8 @@ after_success:
 - docker exec notes /edx/app/notes/notes/.travis/run_coverage.sh
 - codecov
 env:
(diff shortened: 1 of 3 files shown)


morenol:
I think that we don't need to change the Dockerfile, the docker container in the tests should have the needed python interpreters, the thing is that when we use tox -e django22 it seems that is using the default python interpreter, so I think that we just have to call that with py38-django22 just like in https://github.com/edx/course-discovery/blob/59fdec2aed3c15bba4f97bfb2aac416228556ccf/.travis.yml#L37

PR #171, opened 2020-06-11
Adding py3.8 support.

awais786:
https://openedx.atlassian.net/browse/BOM-1628
Adding py3.8 support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,4 @@
 language: python
-python:
-- '3.6'
 branches:
   only:
   - master
@@ -18,6 +16,13 @@ after_success:
 - pip install -U codecov
 - docker exec notes /edx/app/notes/notes/.travis/run_coverage.sh
 - codecov
+
+matrix:
+  - python: 3.8
+    env: TOXENV=py38-django22
+  - python: 3.6
+    env: TOXENV=py36-django22
+
 env:
   global:
   - DOCKER_USERNAME=edxbuilder
--- a/.travis/docker-compose-travis.yml
+++ b/.travis/docker-compose-travis.yml
@@ -16,7 +16,7 @@ services:
     container_name: es
   notes:
     # Uncomment this line to use the official course-discovery base image
-    image: edxops/notes:latest
+    image: openedx/edx-notes-api:latest
 
     # Uncomment the next two lines to build from a local configuration repo
     # build: ../configuration/docker/build/discovery/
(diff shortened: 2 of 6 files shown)


.travis.yml
```diff
@@ -19,6 +19,8 @@ after_success:
 - docker exec notes /edx/app/notes/notes/.travis/run_coverage.sh
 - codecov
 env:
+  - TOXENV=py38-django22
```
adzuci, line 22: Instead of adding a TOXENV here, could you add 3.8 to the python block above like this? https://github.com/edx/demographics/pull/29/files

morenol: Hi @adzuci, the problem with that is that all the tests are run inside of the https://hub.docker.com/r/edxops/notes and it ignores the python version of the travis worker

awais786: @adzuci I did that earlier but tests were running with py3.5 instead of py3.8.

adzuci: Ah, I see, if we install Python 3.8 in the dockerfile will that help, cc @tuchfarber

morenol: @adzuci  the docker image used in the tests is created in the configuration repository and it already has [python3.8](https://github.com/edx/configuration/pull/5791).

The problem is that the default python in that image is python3.5

jmbowman: Also note that pretty soon we'll want to add tests for Django 3.0 also, and that can only be accomplished via setting `TOXENV` appropriately and passing it through all necessary Docker and Makefile layers.

awais786:
@adzuci @nadeemshahzad Please review.

awais786:
@adamstankiewicz @nadeemshahzad Please review.

awais786:
@jmbowman please review.

edx-deployment:
travis has started building this code into a docker file.  Check status at https://travis-ci.org/edx/edx-notes-api/builds/703597235

edx-deployment:
travis has started building this code into a docker file.  Check status at https://travis-ci.org/edx/edx-notes-api/builds/703597235

edx-deployment:
A docker container including this PR has been built and shipped to docker hub.  Check it out at https://hub.docker.com/r/openedx/edx-notes-api/tags?page=1&name=5769196c5d1e06f54053934ad9831644bb1f866b

edx-deployment:
A docker container including this PR has been built and shipped to docker hub.  Check it out at https://hub.docker.com/r/openedx/edx-notes-api/tags?page=1&name=5769196c5d1e06f54053934ad9831644bb1f866b

PR #172, opened 2020-06-18
add python3.8 to Dockerfile (DOS-1246)

nadeemshahzad:


Dockerfile
```diff
@@ -17,11 +17,21 @@ MAINTAINER devops@edx.org
...
+
+RUN apt-get update &&\
+    apt-get install -y locales &&\
+    apt-get install -y locales software-properties-common &&\
+    locale-gen en_US.UTF-8
+
+# Add the deadsnakes PPA to install Python 3.8
+RUN apt-add-repository -y ppa:deadsnakes/ppa
+
+RUN apt-get update && apt-get upgrade -qy && apt-get install language-pack-en locales git python3.6 python3-pip libmysqlclient-dev libssl-dev python3-dev -qy && \
```
adzuci, line 31: I think we may want to leave all apt-get installs on one RUN command to keep the image size smaller and split each package into a new line to make future diffs more readable?

Dockerfile
```diff
@@ -17,11 +17,21 @@ MAINTAINER devops@edx.org
...
-pip3 install --upgrade pip setuptools && \
-rm -rf /var/lib/apt/lists/*
+
+RUN apt-get update &&\
+    apt-get install -y locales &&\
+    apt-get install -y locales software-properties-common &&\
+    locale-gen en_US.UTF-8
+
+# Add the deadsnakes PPA to install Python 3.8
+RUN apt-add-repository -y ppa:deadsnakes/ppa
```
adzuci, line 29: Could we move this up above the first apt-get?

Dockerfile
```diff
@@ -17,11 +17,21 @@ MAINTAINER devops@edx.org
...
 # MySQL-python for performance gains.
-
+# software-properties-common; to get apt-add-repository
 # If you add a package here please include a comment above describing what it is used for
-RUN apt-get update && apt-get upgrade -qy && apt-get install language-pack-en locales git python3.5 python3-pip libmysqlclient-dev libssl-dev python3-dev -qy && \
-pip3 install --upgrade pip setuptools && \
-rm -rf /var/lib/apt/lists/*
+
+RUN apt-get update &&\
+    apt-get install -y locales &&\
```
adzuci, line 24: Could you make the end of line &&\ or && \ consistent across the whole file?

Dockerfile
nadeemshahzad: We need software-properties-common to get apt-add-repository that we can use to add deadsnakes ppa. This is I need a separate RUN for installation.

Dockerfile
nadeemshahzad: Need to install software-properties-common first before using ```apt-add-repository````

Dockerfile
nadeemshahzad: Addressed

Dockerfile
adzuci: That makes sense.  I think we should eventually run build separate images for each version of Python, but for now this seems ok.  Could we include `&&  rm -rf /var/lib/apt/lists/*` at the end of the `RUN apt-get` instructions to clean up the apt cache so it isn’t stored in the layer. See more in the Docker Docs under APT here: https://docs.docker.com/develop/develop-images/dockerfile_best-practices

Another option would be to squash all of the apt layers into one by &&'ing line 23-34 into one run line so we don't store the intermediary layers.

nadeemshahzad: @adzuci Combining multiple RUN statements image size is now ```820MB```

edx-deployment:
travis has started building this code into a docker file.  Check status at https://travis-ci.org/edx/edx-notes-api/builds/701173552

edx-deployment:
A docker container including this PR has been built and shipped to docker hub.  Check it out at https://hub.docker.com/r/openedx/edx-notes-api/tags?page=1&name=b9b2a6d77de8a982815b2e6d4f4b1bcc31758976

PR #178, opened 2020-07-02
 Modifying .travis.yml to more match psudo-standard 

jinder1s:
Changed how the test matrix is created by specifying confi in python and env tag
Added django30 to testing

I'm calling it psuedo-standard cause it has not been agreed upon, its just the norm among our repos.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,37 +1,42 @@
 language: python
+
+python:
+  - 3.5
+  - 3.8
+
+
 branches:
   only:
   - master
 services:
-- docker
+  - docker
 sudo: required
 before_install:
-- make travis_up
+  - make travis_up
 install:
-- pip install -r requirements/travis.txt
+  - pip install -r requirements/travis.txt
 script:
-- make travis_test
-- make travis_pii_check
+  - make travis_test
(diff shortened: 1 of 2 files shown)


morenol:
Hi @jinder1s, if we apply these changes, internally all the tests will be run in python3.6

You can check this in the output of the tests:
```
============================= test session starts ==============================
platform linux -- Python 3.6.9, pytest-5.4.3, py-1.9.0, pluggy-0.13.1
```

jinder1s:
@morenol thanks for pointing out the python 3.6.9 issue. Question, do you know why is that happening? Is this cause this is using docker?

morenol:
>  Question, do you know why is that happening? Is this cause this is using docker?

Yes, it is due to it using the docker image so it ignores the python version of the travis worker. It is using the default python interpreter in the docker container.

 For that reason we had to use 
        `TOXENV=py38-django22`

jinder1s:
okay, makes sense, thanks for the info!

jinder1s:
Closing this PR since this repo's travis testing is different from the usual, thus the diverging format.

PR #185, opened 2020-08-04
deploy docker image on merge to master

syed-awais-ali:


.github/workflows/push-docker-image.yml
```diff
@@ -0,0 +1,17 @@
...
+  push:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push'
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Build and Push docker image
+        run : ./.travis/deployment_prs.sh
```
adzuci, line 17: Hmm, @syed-awais-ali I took a look at https://github.com/edx/course-discovery/pull/2655/files#diff-b67911656ef5d18c4ae36cb6741b7965R119 and realize that we call a make command directly in course-discovery, could we make notes follow the same pattern and pull the .travis script logic into the top level makefile so that new eyes can understand the logic more easily?

.github/workflows/deployment_prs.sh
```diff
@@ -19,30 +19,30 @@ cd -
 make travis_docker_push
```
adzuci, line 19: @syed-awais-ali One last thing, could you update this to point at `docker_push`?

PR #247, opened 2021-01-06
upgrade gunicorn

Ali-D-Akbar:
[PROD-333](https://openedx.atlassian.net/browse/PROD-333)

Upgrading gunicorn above 19.5.0 which resolved potential vulnerability in process_headers.


fredsmith (approved):

adzuci (approved):
Feel free to ignore the Trivy error if we want to ship this quickly, I will fix it in a separate PR and if there are any findings related to these change I will make sure they are addressed.

PR #251, opened 2021-02-04
Add common constraint

iamsobanjaved:
JIRA: https://openedx.atlassian.net/browse/BOM-2257

### DIFF
--- a/requirements/travis.txt
+++ b/requirements/travis.txt
@@ -4,22 +4,47 @@
 #
 #    make upgrade
 #
-appdirs==1.4.4            # via virtualenv
-certifi==2020.12.5        # via requests
-chardet==4.0.0            # via requests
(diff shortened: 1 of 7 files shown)


requirements/travis.txt
```diff
@@ -4,22 +4,47 @@
 #
 #    make upgrade
```
adzuci, line 5: Do we still run tests in Travis / should we file a ticket to update that?

iamsobanjaved: It's being used in [docker-compose-travis.yml](https://github.com/edx/edx-notes-api/blob/master/.travis/docker-compose-travis.yml)

iamsobanjaved: Should I merge this PR?

PR #293, opened 2022-04-05
fix: modernize requirements

mikix:
- Directly reference common_constraints from github rather than making a local copy and editing it - none of our edits are needed anymore and common_constraints feels conservative enough these days that I don't think we need this capability.
- Run make upgrade to update a bunch of dependencies.


Makefile
```diff
@@ -65,29 +65,13 @@ develop: test.requirements
 piptools: ## install pinned version of pip-compile and pip-sync
 	pip install -r requirements/pip-tools.txt
 
-COMMON_CONSTRAINTS_TXT=requirements/common_constraints.txt
-.PHONY: $(COMMON_CONSTRAINTS_TXT)
-$(COMMON_CONSTRAINTS_TXT):
-	wget -O "$(@)" https://raw.githubusercontent.com/edx/edx-lint/master/edx_lint/files/common_constraints.txt || touch "$(@)"
```
mikix, line 71: One reason I prefer to just directly reference this file rather than download and update it when running `make upgrade` is that the base docker images for notes don't have wget. So we end up hitting the `|| touch` path, and local engineers trying to run make upgrade may have out of date constraints.

Makefile
```diff
@@ -65,29 +65,13 @@ develop: test.requirements
...
-
 upgrade: export CUSTOM_COMPILE_COMMAND=make upgrade
-upgrade: $(COMMON_CONSTRAINTS_TXT) piptools ## update the requirements/*.txt files with the latest packages satisfying requirements/*.in
-	sed 's/pyjwt\[crypto\]<2.0.0//g' requirements/common_constraints.txt > requirements/common_constraints.tmp
```
cdeery, line 75: is this flushing out old settings? Not sure what's going on here

cdeery: None of these settings are in common constraints anymore, right?

cdeery (approved):
Thanks,

Looks good

Makefile
mikix: Exactly. They aren’t in common constraints anymore. I get the impression that common constraints was a bit more aggressive / platform-focused in past? But I don’t remember the history.

PR #335, opened 2023-05-31
docs: Explain why CORS setting is permissive

timmc-edx:
I'm not familiar with this IDA, so hopefully this is correct!


mattcarter (approved):
LGTM 👍🏻

PR #339, opened 2023-07-06
chore: install pkg-config for PyMySQL compatibility

leangseu-edx:
https://github.com/edx/edx-arch-experiments/issues/349

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,6 +26,7 @@ RUN apt-get update && \
     locales \
     git \
     libmysqlclient-dev \
+    pkg-config \
     libssl-dev \
     build-essential \
     python3.8-dev \


timmc-edx:
Same recommendations here about `DEBIAN_FRONTEND=noninteractive` and/or `--no-install-recommends` being needed, and optionally a comment describing which packages are needed for what.

leangseu-edx:
> Same recommendations here about `DEBIAN_FRONTEND=noninteractive` and/or `--no-install-recommends` being needed, and optionally a comment describing which packages are needed for what.

I built this dockerfile. It seems to work without those option.

timmc-edx (approved):
Hmm, interesting! I'd like to understand why it's a problem for some dockerfiles and not others. But if it works, that's good enough for me. :-)

mattcarter (approved):
if the container runs correctly w/o the config changes,  LGTM 👍🏻

PR #370, opened 2023-11-16
fix: remove new relic pin to unblock django upgrade

rgraber:
Unpins newrelic to bring in a bug fix that is causing 404s to be converted to 500, which is blocking the Django 4.2 upgrade.


matthugs (approved):

mattcarter (approved):
approved

PR #392, opened 2024-02-23
Add support for Python 3.12

irtazaakram:
ticket: https://github.com/openedx/axim-engineering/issues/1059


notesapi/v1/tests/test_views.py
```diff
@@ -387,9 +388,14 @@ def test_read_all_no_annotations(self):
         headers["course_id"] = "a/b/c"
         response = self.client.get(reverse('api:v1:annotations'), headers)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
-        self.assertDictContainsSubset(
-            {'total': 0, 'rows': []}, response.data, "no annotation should be returned in response"
-        )
+        if sys.version_info[1] == 12:  # python 3.12
+            self.assertEqual(
```
feanil, line 392: AssertEqual has always been present, why not just use that in both versions and not have a conditional?

feanil (changes requested):
Also, don't forget to update the classifiers in setup.py

notesapi/v1/tests/test_views.py
irtazaakram: With python 3.8 we cannot use `|` operator with dicts
`E       TypeError: unsupported operand type(s) for |: 'dict' and 'dict'`

It was added in python 3.9 https://peps.python.org/pep-0584/
Error https://github.com/openedx/edx-notes-api/actions/runs/8018877609/job/21905671987#step:6:58

irtazaakram: Also there's no setup.py file in this repo.

feanil: gotcha, thanks for the  context.

requirements/base.txt
```diff
@@ -6,7 +6,7 @@
 #
 asgiref==3.7.2
     # via django
-backports-zoneinfo==0.2.1
+backports-zoneinfo==0.2.1; python_version < '3.9'
```
feanil, line 9: Did the python version restriction happen because of of an update to a constraint or input file? or because we ran `make upgrade` with a newer python version? or something else?

irtazaakram: It was supposed to be part of `constraints.txt` which I missed. I've added it here now https://github.com/openedx/edx-notes-api/pull/392/commits/a388a846e38bc6308dffb7724028ac47b4001404

feanil (approved):

PR #407, opened 2024-05-02
fix: build and push docker images CI 

irtazaakram:
- Resolves https://github.com/openedx/edx-notes-api/issues/405

Updated script as per official docs https://github.com/docker/build-push-action

### DIFF
--- a/.github/workflows/push-docker-image.yml
+++ b/.github/workflows/push-docker-image.yml
@@ -5,21 +5,19 @@ on:
     branches:
       - master
       - open-release/**
+
 jobs:
   push:
     runs-on: ubuntu-latest
     if: github.event_name == 'push'
 
     steps:
-      - name: Checkout
-        uses: actions/checkout@v3
-
       # Use the release name as the image tag if we're building an open release tag.
       # Examples: if we're building 'open-release/olive.master', tag the image as 'olive.master'.
       # Otherwise, we must be building from a push to master, so use 'latest'.
       - name: Get tag name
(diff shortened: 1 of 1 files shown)


.github/workflows/push-docker-image.yml
```diff
@@ -28,37 +26,34 @@ jobs:
...
-      #   with:
-      #     push: true
-      #     username: ${{ secrets.DOCKERHUB_USERNAME }}
-      #     password: ${{ secrets.DOCKERHUB_PASSWORD }}
-      #     target: prod
-      #     repository: openedx/edx-notes-api
-      #     tags: ${{ steps.get-tag-name.outputs.result }},${{ github.sha }}
-      #     platforms: linux/amd64,linux/arm64
\ No newline at end of file
+      - name: Build and push Prod Docker image
```
farhan, line 51: @feanil 
Prod images were commented in this [PR](https://github.com/openedx/edx-notes-api/pull/320/files#diff-17cd2e03fb1dfb62371b2743de71db3bf8e2ab0a5fd0016ae5f8dc31ee690fd8R41
)
Do you have information we did we do that?
Seems ok to me to uncomment it.

farhan (changes requested):

.github/workflows/push-docker-image.yml
feanil: If we're not pushing the images, then we don't need this workflow at all.  Let's just remove it, it was 2U specific anyway.

PR #476, opened 2025-03-07
fix: Use edx-django-utils API instead of calling newrelic directly

timmc-edx:
These changes will cause telemetry to start being reported when annotations are created. (No change for whether notesserver heartbeat calls emit telemetry, as that is currently only supported for New Relic anyhow.)

This also removes the newrelic dependency, though it is still pulled in by edx-django-utils indirectly, for now.

[EDIT: This originally just wrapped the newrelic import in a try block, but the new PR removes newrelic references entirely.]

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -109,9 +109,7 @@ meilisearch==0.34.0
 mysqlclient==2.2.7
     # via -r requirements/base.in
 newrelic==10.7.0
-    # via
-    #   -r requirements/base.in
-    #   edx-django-utils
+    # via edx-django-utils
 packaging==24.2
     # via
     #   django-nine
--- a/notesapi/v1/views/common.py
+++ b/notesapi/v1/views/common.py
@@ -1,7 +1,6 @@
 #  pylint:disable=possibly-used-before-assignment
 import json
 import logging
-import newrelic.agent
 from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.db.models import Q
@@ -11,6 +10,11 @@ from rest_framework import status
 from rest_framework.generics import GenericAPIView, ListAPIView
 from rest_framework.response import Response
 from rest_framework.views import APIView
+
+try:
+    import newrelic.agent
+except ImportError:  # pragma: no cover
+    newrelic = None  # pylint: disable=invalid-name
 
 from notesapi.v1.models import Note
 from notesapi.v1.serializers import NoteSerializer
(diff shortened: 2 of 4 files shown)


nsprenkle (approved):

timmc-edx:
Instead of `newrelic.agent.add_custom_parameter` you'd want to call `edx_django_utils.monitoring.set_custom_attribute`. Monitoring utils docs: https://github.com/openedx/edx-django-utils/tree/master/edx_django_utils/monitoring

I was just going for the least-work option, but it does look like edx-django-utils is already in the IDA's requirements, so maybe it would be better to just go ahead and make that change. Either way works for me, I'm just clearing the path for a DEPR. :-)

timmc-edx:
I went back and made the better fix -- PR name and description updated, too.

requirements/base.txt
```diff
@@ -180,7 +178,6 @@ stevedore==5.4.1
 typing-extensions==4.12.2
     # via
     #   edx-opaque-keys
-    #   elasticsearch-dsl
```
timmc-edx, line 183: Author's note: I'm not sure why this transitive dependency comment was removed when I ran `make compile-requirements` locally, but it shouldn't have any effect. (Might be a Mac vs. Linux difference, or a leftover of someone manually editing the files at some point.)

requirements/base.txt
```diff
@@ -109,9 +109,7 @@ meilisearch==0.34.0
 mysqlclient==2.2.7
     # via -r requirements/base.in
 newrelic==10.7.0
```
nsprenkle, line 111: Looks like newrelic is still included in the requirements, is that intentional?

nsprenkle (approved):

requirements/base.txt
timmc-edx: Yes -- if you look at the comments below that line, you can see that it's still being pulled in via edx-django-utils (for now).

sarina: What's the future plan for removing it from requirements?

timmc-edx: As far as I can tell, this was the last instance in the edx and openedx orgs where we still had a direct `import newrelic` -- so after this merges, I think we're clear to remove `newrelic` from the edx-django-utils `base.in` as well. I don't have a specific plan for that, but I did want to clear the path for it.

ormsbee (approved):

sarina (approved):

requirements/base.txt
timmc-edx: Update: Filed a DEPR for this: https://github.com/openedx/public-engineering/issues/360

